PR #5, opened 2015-11-24
KAFKA-2881: Organized Consumer APIs and Configs

guozhangwang:


guozhangwang:
Ping @gwenshap @junrao for reviews.

gwenshap:
The patch looks good, but I'm concerned that by committing to kafka-site directly, we will lose sync with the docs in the trunk.

Mind making this a kafka trunk PR and then we can push the change to both repos?

guozhangwang:
Good point, will do that.

PR #20, opened 2016-09-30
new design

derrickdoo:
Updated theme 
- new nav and layout
- updated homepage copy
- mobile styles


0100/protocol.html
```diff
@@ -16,208 +16,215 @@
 -->
 
 <!--#include virtual="../includes/header.html" -->
+<!--#include virtual="../includes/top.html" -->
+<div class="content">
```
gwenshap, line 20: it doesn't look like we are closing this div tag anywhere?

0100/protocol.html
```diff
@@ -16,208 +16,215 @@
...
+	<!--#include virtual="../includes/nav.html" -->
+	<div class="right">
```
gwenshap, line 22: couldn't find where we are closing this either?

derrickdoo: Both of these get closed in /includes/footer.html. The old /includes/header.html used to leave these tags open and they'd get closed in footer. 

I broke header.html up a little bit so we could have more flexibility with placing page specific stuff on top of .content.

gwenshap:
LGTM

PR #24, opened 2016-10-11
update htaccess to load images nested inside of doc folders

derrickdoo:


derrickdoo:
- better formatting for code elements
- don't rewrite requests for images nested in the specific doc version directories

hachikuji:
LGTM. Thanks for the fixes!

PR #35, opened 2016-12-07
KAFKA-4498: [DO NOT MERGE] add rewrite rule for streams

guozhangwang:
For @derrickdoo to review and take over.

Add the rule to avoid rewrite with [version]/documentation#streams_[any]

Add the rule to rewrite [version]/documentation#streams_[any] to [version]/documentation/streams#[any]
Add the rule to rewrite /documentation#streams_[any] to /documentation/streams#[any]


guozhangwang:
Actually thinking about it a bit more, for versioned address we could still apply the rule for all versions. And any old versions that does not have the streams section, the hashtag will never match anyways.

PR #36, opened 2016-12-08
Streams standalone docs

derrickdoo:
Here's the changes we talked about to pull Streams documentation out to it's own page. Ended up having to do the redirect on URLs that contain anchor tags that match the pattern 'streams_' with JS in the head section of the page because anchor tags aren't available on the server so you can't do rewrite rules based off them.

(also added in a change I thought was already out there to show info about Kafka Summit on the events page)


derrickdoo:
@guozhangwang here's a PR for the standalone Streams docs page

guozhangwang:
Thanks @derrickdoo , LGTM overall. One meta question:

Does any of those changes need to be in `kafka`'s docs folder as well? During each release we will do a copy-paste from the built tars to `kafka-site`: 

https://cwiki.apache.org/confluence/display/KAFKA/Release+Process#ReleaseProcess-Websiteupdateprocess

Next time when we do this, some of the changes may be reverted.

So maybe we need another PR on the `kafka` repo as well for some of those changes, other changes like the `event.html` are not included in the docs folder and hence we do not need them in the `kafka` repo.

derrickdoo:
@guozhangwang yes, 2 things need to happen to new copies of stuff in side the versioned docs folder. 

1. streams.html will need to have the full shell around it, so the includes at the top and bottom as seen in 0101/streams.html need to be there all the time now.

2. paths to images in all the new doc files (0101 and beyond) need to include the full path to the image directory.

I'll make a PR to the kafka repo to make sure everything is all set for the next release

guozhangwang:
Thanks @derrickdoo ! Please add me to reviewers list when you created the PR.

guozhangwang:
Merged, thanks!

PR #52, opened 2017-04-06
SMT example, correct classnames in the generated transformations list…

gwenshap:
… and more Connect in TOC

### DIFF
--- a/0102/connect.html
+++ b/0102/connect.html
@@ -102,7 +102,7 @@
 
     <h4><a id="connect_transforms" href="#connect_transforms">Transformations</a></h4>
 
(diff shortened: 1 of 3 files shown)


gwenshap:
@ewencp ?

ewencp (approved):
@gwenshap This is just an update from what's in Kafka's 0.10.2, right? Not even sure it requires a PR since it's a mechanical update.

In any case, LGTM so you can feel free to merge.

gwenshap:
Ismael said I should get those mechanical thingies reviewed just in case I'm missing files or stuff.. Thanks!

PR #58, opened 2017-05-30
Break apart Streams docs

derrickdoo:
- Break apart Streams docs into 4 sections


includes/_footer.htm
```diff
@@ -32,7 +32,8 @@
 				'streams',
 				'quickstart',
 				'toc',
-				'upgrade'
+				'upgrade',
+				'content'
```
guozhangwang, line 36: I found this name a bit misleading.. could we rename it to `streams-content` or `streams-subpages`?

guozhangwang (commented):
One minor comment, otherwise LGTM.

includes/_footer.htm
derrickdoo: I did this on purpose so we could have a generic name for containers that we know aren't going to rendered alongside themselves. It just makes it easier for other to copy and paste existing pages and build off of them.

guozhangwang: Got it, makes sense.

guozhangwang:
Merged to `asf-site`.

PR #59, opened 2017-05-31
Add auto syntax highlighter

derrickdoo:
Added plugin to do auto syntax highlighting for code snippets in our docs. (https://github.com/syntaxhighlighter/syntaxhighlighter/wiki)

Here's an example of how to syntax highlight some java code:
`<pre class="brush: java;">your code goes here</pre>`

Here's what this would look like:
image


guozhangwang:
LGTM. Merged to asf-site.

PR #60, opened 2017-06-08
Update delivery semantics section for KIP-98

hachikuji:


hachikuji:
cc @guozhangwang @apurva @ijuma

0110/design.html
```diff
@@ -261,15 +262,23 @@
...
-    </ol>
     <p>
-    So effectively Kafka guarantees at-least-once delivery by default and allows the user to implement at most once delivery by disabling retries on the producer and committing its offset prior to processing a batch of
-    messages. Exactly-once delivery requires co-operation with the destination storage system but Kafka provides the offset which makes implementing this straight-forward.
+    A special case is when the output system is just another Kafka topic (e.g. in a Kafka Streams application). Here we can leverage the new transactional producer capabilities in 0.11.0.0 that were mentioned above.
+    Since the consumer's position is stored as a message in a topic, we can ensure that that topic is included in the same transaction as the output topics receiving the processed data. If the transaction is aborted,
+    the consumer's position will revert to its old value and none of the output data will be visible to consumers. To enable this, consumers support an "isolation level" to achieve this. In the default
```
ijuma: Seems like this sentence could be clarified a little. `we can ensure that that topic is included in the same transaction`, `that` is repeated and it's unclear what it means to include a topic in a transaction. Did you mean the updates to the topic or something along those lines?

Also `the consumer's position will revert to its old value and none of the output data will be visible to consumers`. It may be worth qualifying `consumer` and `consumers` in that sentence.

0110/design.html
```diff
@@ -261,15 +262,23 @@
...
+    "read_uncommitted" mode, all messages are visible to consumers even if they were part of an aborted transaction, but in "read_committed" mode, the consumer will only return data from transactions which were committed
+    (and any messages which were not part of any transaction).
+    <p>
+    So effectively Kafka guarantees at-least-once delivery by default, and allows the user to implement at-most-once delivery by disabling retries on the producer and committing its offset prior to processing a batch of
```
ijuma: I wonder if we should mention the stronger guarantees first.

ijuma (approved):
Thanks for the PR. LGTM. Just a couple of comments/suggestions.

0110/design.html
hachikuji: Good point.

0110/design.html
hachikuji: Thanks. I will try to clarify these lines.

0110/design.html
guozhangwang: +1. Maybe we could say "Kafka guarantees exactly-once delivery, but default it is turned off to optimize performance with at-least-once delivery" blah blah.

0110/design.html
```diff
@@ -261,15 +262,23 @@
...
+    messages. Exactly-once delivery is supported when processing messages between Kafka topics, such as in Kafka Streams applications. Exactly-once delivery for other destination storage system generally requires
```
guozhangwang: By just reading the first sentence it looks like "exactly-once is constraint to only the case when processing messages between Kafka topics" but only the second line points out that e.g. even with an external system like HDFS, Kafka Connect can also achieve EOS by coordinating with the system.

Maybe we can also re-phrase it as "Exactly-once delivery is supported out-of-the-box when ...; for other destination storage system it is generally required to cooperate ... to achieve exactly-once".

guozhangwang: nit: add a ref link when mentioning Kafka Streams? https://kafka.apache.org/documentation/streams

apurvam (approved):
I looked at this rather late, but the current version looks good to me!

hachikuji:
@guozhangwang @ijuma @apurvam Pushed a couple changes based on the review. Please take another look.

guozhangwang:
lgtm.

0110/design.html
```diff
@@ -264,21 +264,22 @@
...
+    same transaction as the output topics receiving the processed data. If the transaction is aborted, the consumer's position will revert to its old value and none of the output data will be visible to consumers. Consumers support an "isolation level" configuration
+    to achieve this. In the default "read_uncommitted" mode, all messages are visible to consumers even if they were part of an aborted transaction, but in "read_committed" mode, the consumer will only return data from
+    transactions which were committed (and any messages which were not part of any transaction).
+    <p>
+    When writing to an external system, the limitation is in the need to coordinate the consumer's position with what is actually stored as output. The classic way of achieving this would be to introduce a two-phase
+    commit between the storage for the consumer position and the storage of the consumers output. But this can be handled more simply and generally by simply letting the consumer store its offset in the same place as
```
ijuma: "storage of the consumer position"? Since we talk about consumer position, should we be saying `consumer's output`?

0110/design.html
```diff
@@ -264,21 +264,22 @@
...
+    its output. This is better because many of the output systems a consumer might want to write to will not support a two-phase commit. As an example of this, our Hadoop ETL that populates data in HDFS stores its
```
ijuma: `our ETL` doesn't seem right because it's a Confluent Connector right?

0110/design.html
```diff
@@ -264,21 +264,22 @@
...
+    offsets in HDFS with the data it reads so that it is guaranteed that either data and offsets are both updated or neither is. We follow similar patterns for many other data systems which require these stronger
+    semantics and for which the messages do not have a primary key to allow for deduplication.
+    <p>
+    So effectively Kafka supports exactly-once delivery in <a href="https://kafka.apache.org/documentation/streams">Kafka Streams</a>, and the transactional producer/consumer can be used generally to provide
+    exactly-once delivery when transfering and processing data between Kafka topics. Exactly-once delivery for other destination storage system generally requires cooperation with that system, but Kafka provides the
```
ijuma: "for other destination systems generally requires cooperation such such systems" maybe

0110/design.html
```diff
@@ -264,21 +264,22 @@
...
+    offset which makes implementing this straight-forward. Otherwise, Kafka guarantees at-least-once delivery by default, and allows the user to implement at-most-once delivery by disabling retries on the producer and
```
ijuma: A bit optimistic: "straight-forward" :)

ijuma (commented):
Some nits, looks good otherwise.

0110/design.html
hachikuji: Haha, this was the original language, but I raised an eyebrow when I read that as well. Maybe "feasible" is nearer the mark.

0110/design.html
hachikuji: I'm not sure I understand the second question. Can you elaborate?

hachikuji:
I'm going to reopen this as a PR against kafka since that is where design.html is maintained.

PR #63, opened 2017-06-23
Do not use hyphens in wrapping property names in tables

derrickdoo:
Hyphenating long property names that wrap inside of tables cells is causing confusion. CSS to prevent this from happening. (mentioned in KAFKA-5460)


guozhangwang:
LGTM. Merged to asf-site.

PR #64, opened 2017-07-06
MINOR: Add release dates to downloads page

ewencp:


ewencp:
@ijuma @guozhangwang @junrao I think it'd be good to start including these to help indicate to users how old the releases they are downloading are.

guozhangwang:
What about pre-0.8.2 releases?

ewencp:
I can dig up the release dates, I just grabbed the ones that were readily available via the git history for this file (which, admittedly, might be slightly off). I'm happy to round them out via svn history and/or digging through mailing list history if people think it is worth it.

ijuma:
Good idea. As you said, the dates seem to differ slightly from: https://issues.apache.org/jira/browse/KAFKA/?selectedTab=com.atlassian.jira.jira-projects-plugin:versions-panel

It would be nice to get pre-0.8.2 dates since we're doing this.

ewencp:
I grabbed the dates from that page and filled in the beta ones based on mailing list posts. The actual dates seem to be inconsistent and the JIRA release dates aren't always maintained (I know because I dug up dates and closed a few other releases when I last ran one), but tbh the point here is just the rough timing anyway.

ijuma (approved):
Thanks for posting all of the dates and updating JIRA too. Nice to be able to see it all in one place. LGTM. I will let you merge it.

PR #65, opened 2017-07-06
MINOR: add Interactive Queries docs

dguy:


dguy:
@guozhangwang

guozhangwang:
Actually I think we do not need a PR for this: as a committer anyone can do the docs jar packaging -> pasting to `kafka-site` and push to `asf-site` branch. You can follow this step of the release process here:

https://cwiki.apache.org/confluence/display/KAFKA/Release+Process#ReleaseProcess-Websiteupdateprocess

dguy:
Can't i just merge this PR? The process there seems more about creating a new version for a release, i.e, i don't want to update everything

guozhangwang:
Generally speaking separate commits on `kafka` and `kafka-site` may have the risk of diverge on these two repos if people forget to make the corresponding changes to `kafka-site` while pushed to `kafka` etc. 

Re-package the docs to `kafka-site` won't update everything as long as it is from the corresponding branch (i.e. 0.11.0 in this case), the affected files should only be those gets modified by the patch on `kafka`.

dguy:
@guozhangwang , yep had the discussion with @ijuma today. Managed to break everything in the process due to changes that weren't in `kafka` that were in `kafka-site` and changes in `kafka` that were broken. Anyway, it was fun! ;-)

PR #74, opened 2017-09-05
Added portoseguro, micronauticsresearch & cj logos to powered by page

manjuapu:


powered-by.html
```diff
@@ -412,7 +412,22 @@
             "logo": "rabobank.jpg",
             "logoBgColor": "#ffffff",
             "description": "Rabobank is one of the 3 largest banks in the Netherlands. Its digital nervous system, the Business Event Bus, is powered by Apache Kafka and Kafka Streams."
-        }
+        },{
+    "link":  "http://www.portoseguro.com.br/",
```
ewencp: Nit: can we make indentation match the above for consistency? If we don't do this we tend to get people's editors reformatting things which adds noise to the diffs

ewencp (commented):
Mostly looks good. Can we get rid of the `.DS_Store` files? We generally don't want to check those in. Otherwise, just a minor formatting diff, but the changes to the page look good.

ewencp (approved):
Looks like one `.DS_Store` is still in there, but I'll just get it on commit. LGTM, thanks!

PR #90, opened 2017-10-05
Kafka registered trademark

derrickdoo:
Update art assets and content to reflect registered status of the Kafka trademark


junrao:
@derrickdoo : Thanks for the patch. LGTM

PR #133, opened 2018-03-06
Update for 1.0.1 release

ewencp:


10/documentation/streams/architecture.html
```diff
@@ -15,5 +15,5 @@
  limitations under the License.
 -->
 
-<!-- should always link the latest release's documentation -->
+<!-- should always link the the latest release's documentation -->
```
guozhangwang, line 18: duplicate the，ditto below

guozhangwang (approved):
nit comment，lgtm

10/documentation/streams/architecture.html
ewencp: this is just auto-staged docs via `release.py` from upstream, this only goes through a PR to satisfy the new requirements for PR review before pushing. this would need to be fixed [at the source](https://github.com/apache/kafka/blob/1.0/docs/documentation/streams/architecture.html#L18). The fact that this is showing up as a diff implies this was incorrectly fixed in this repo but not upstream for this branch.

PR #148, opened 2018-07-03
MINOR: update download links

mjsax:


guozhangwang:
What's the intention of using `https`, and why changing to `https://archive.apache.org`?

mjsax:
`https` should always be the default IMHO (cf https://issues.apache.org/jira/browse/KAFKA-6222). For the link change compare: https://issues.apache.org/jira/browse/KAFKA-6223

guozhangwang:
> https should always be the default IMHO (cf https://issues.apache.org/jira/browse/KAFKA-6222). For the link change compare: https://issues.apache.org/jira/browse/KAFKA-6223

ack.

guozhangwang (approved):

lindong28:
LGTM. Verified that this patch only achieves the links for non-latest versions.

lindong28 (approved):

PR #149, opened 2018-07-08
Update docs for 1.0.2 release

mjsax:


10/upgrade.html
```diff
@@ -129,17 +129,77 @@ <h5><a id="upgrade_100_new_protocols" href="#upgrade_100_new_protocols">New Prot
          be used if the SaslHandshake request version is greater than 0. </li>
 </ul>
 
-<h5><a id="upgrade_100_streams" href="#upgrade_100_streams">Upgrading a 1.0.0 Kafka Streams Application</a></h5>
+<h5><a id="upgrade_100_streams_from_0110" href="#upgrade_100_streams_from_0110">Upgrading a 0.11.0 Kafka Streams Application</a></h5>
 <ul>
     <li> Upgrading your Streams application from 0.11.0 to 1.0.0 does not require a broker upgrade.
```
guozhangwang, line 134: line 134 above: to `1.0.2` or `1.0.x` or `1.0` ?

guozhangwang (approved):
One minor comment, otherwise lgtm.

Please feel free to merge afterwards.

10/upgrade.html
mjsax: The whole section is about upgrading to `1.0`. I just fixed the headline, because is says `1.0.0` what is incorrect (upgrading from 1.0.0 to 1.0 does not make sense)

mjsax: @guozhangwang Merging this as-is. If you have further comments, we can do a follow up PR.

guozhangwang: My concern was that, after 1.0.2 is released, the docs under `10` would actually for 1.0.2, so the content mentioning `upgrading to 1.0.0` may be a bit confusing.

Maybe we can just rename all the versions in content to only include up to minor digits, like `1.0`, or say `1.0.x` so that when we release new bug-fix versions we do not need to change the versions any more.

mjsax: Follow ups:
 - https://github.com/apache/kafka/pull/5372
 - https://github.com/apache/kafka-site/pull/151

PR #153, opened 2018-07-19
Update download links for 1.1.1 release

lindong28:


lindong28:
@mjsax @ewencp do you have time to review the patch? Thanks.

downloads.html
```diff
@@ -22,13 +43,13 @@ <h3 class="download-version">1.1.0<a href="#1.1.0"><i class="fas fa-link " style
                 <a href="https://www.apache.org/dist/kafka/1.1.0/RELEASE_NOTES.html">Release Notes</a>
             </li>
             <li>
-                Source download: <a href="https://www.apache.org/dyn/closer.cgi?path=/kafka/1.1.0/kafka-1.1.0-src.tgz">kafka-1.1.0-src.tgz</a> (<a href="https://www.apache.org/dist/kafka/1.1.0/kafka-1.1.0-src.tgz.asc">asc</a>, <a href="https://www.apache.org/dist/kafka/1.1.0/kafka-1.1.0-src.tgz.sha512">sha512</a>)
+                Source download: <a href="https://archive.apache.org/dist/kafka/1.1.0/kafka-1.1.0-src.tgz">kafka-1.1.0-src.tgz</a> (<a href="https://archive.apache.org/dist/kafka/1.1.0/kafka-1.1.0-src.tgz.asc">asc</a>, <a href="https://archive.apache.org/dist/kafka/1.1.0/kafka-1.1.0-src.tgz.sha512">sha512</a>)
```
ewencp, line 46: Are these right? I think we've been told we're not supposed to use the dist URLs and instead use the mirrors.

downloads.html
```diff
@@ -22,13 +43,13 @@ <h3 class="download-version">1.1.0<a href="#1.1.0"><i class="fas fa-link " style
...
             </li>
             <li>
                 Binary downloads:
                 <ul>
-                    <li>Scala 2.11 &nbsp;- <a href="https://www.apache.org/dyn/closer.cgi?path=/kafka/1.1.0/kafka_2.11-1.1.0.tgz">kafka_2.11-1.1.0.tgz</a> (<a href="https://www.apache.org/dist/kafka/1.1.0/kafka_2.11-1.1.0.tgz.asc">asc</a>, <a href="https://www.apache.org/dist/kafka/1.1.0/kafka_2.11-1.1.0.tgz.sha512">sha512</a>)</li>
-                    <li>Scala 2.12 &nbsp;- <a href="https://www.apache.org/dyn/closer.cgi?path=/kafka/1.1.0/kafka_2.12-1.1.0.tgz">kafka_2.12-1.1.0.tgz</a> (<a href="https://www.apache.org/dist/kafka/1.1.0/kafka_2.12-1.1.0.tgz.asc">asc</a>, <a href="https://www.apache.org/dist/kafka/1.1.0/kafka_2.12-1.1.0.tgz.sha512">sha512</a>)</li>
+                    <li>Scala 2.11 &nbsp;- <a href="https://archive.apache.org/dist/kafka/1.1.0/kafka_2.11-1.1.0.tgz">kafka_2.11-1.1.0.tgz</a> (<a href="https://archive.apache.org/dist/kafka/1.1.0/kafka_2.11-1.1.0.tgz.asc">asc</a>, <a href="https://archive.apache.org/dist/kafka/1.1.0/kafka_2.11-1.1.0.tgz.sha512">sha512</a>)</li>
```
guozhangwang, line 51: Could we update the link for 1.0.2 and 0.11.0.3 below as well?

guozhangwang (approved):
Thanks @lindong28 . Left a minor comment otherwise LGTM. Please feel free to merge afterwards.

downloads.html
```diff
@@ -5,12 +5,33 @@
...
+            <li>
+                <a href="https://www.apache.org/dist/kafka/1.1.1/RELEASE_NOTES.html">Release Notes</a>
+            </li>
+            <li>
+                Source download: <a href="https://www.apache.org/dyn/closer.cgi?path=/kafka/1.1.1/kafka-1.1.1-src.tgz">kafka-1.1.1-src.tgz</a> (<a href="https://www.apache.org/dist/kafka/1.1.1/kafka-1.1.1-src.tgz.asc">asc</a>, <a href="https://www.apache.org/dist/kafka/1.1.1/kafka-1.1.1-src.tgz.sha512">sha512</a>)
+            </li>
+            <li>
+                Binary downloads:
+                <ul>
+                    <li>Scala 2.11 &nbsp;- <a href="https://www.apache.org/dyn/closer.cgi?path=/kafka/1.1.1/kafka_2.11-1.1.1.tgz">kafka_2.11-1.1.1.tgz</a> (<a href="https://www.apache.org/dist/kafka/1.1.1/kafka_2.11-1.1.1.tgz.asc">asc</a>, <a href="https://www.apache.org/dist/kafka/1.1.1/kafka_2.11-1.1.1.tgz.sha512">sha512</a>)</li>
```
guozhangwang: Should we use the archive link as well?

lindong28: Hmm... this download link is for 1.1.1 release which is the latest release for 1.1 branch. My understanding should not use archive link here, right?

downloads.html
lindong28: This link was copied and modified based on the download link for 1.0.0 source code which in turn is from https://github.com/apache/kafka-site/pull/148/files.

I am not very sure what is the mirror URL and whether we should use that. It seems like an existing issue here. If we can have quick answer (e.g. a download link to be used here) then I am happy to update it.

downloads.html
lindong28: It looks like 1.0.2 and 0.11.0.3 are both the latest version for the corresponding branch. So it looks like we want to use archive link everywhere. Sure, let me update it.

downloads.html
lindong28: https://issues.apache.org/jira/browse/KAFKA-6222 says that we should use https://www.apache.org/dist/kafka but we should not use dist.apache.org. So it looks like the current download links are OK?

downloads.html
lindong28: Fixed as suggested.

downloads.html
lindong28: Fixed as suggested.

lindong28:
@guozhangwang @ewencp BTW, I am not able to mark the 1.1.1 version as released in https://issues.apache.org/jira/projects/KAFKA?selectedItem=com.atlassian.jira.jira-projects-plugin%3Arelease-page&status=unreleased. This is likely because I am not in the admin list for Kafka JIRA. Can you help mark that version as released?

guozhangwang:
> This is likely because I am not in the admin list for Kafka JIRA. Can you help mark that version as released?

No worries. I can do that.

downloads.html
ewencp: ah, ok, i was just misremembering the exact URLs

lindong28:
@guozhangwang @ewencp Thanks for your help and review.

PR #154, opened 2018-07-19
Update download links for 1.1.1 release followup

lindong28:


lindong28:
@guozhangwang @ewencp It looks like the archive link does not work for 1.1.1 yet. So we need to use the non-achieve link for 1.1.1 release for now. Can you take a look?

guozhangwang (approved):

PR #160, opened 2018-09-01
Make links recognizable

courajs:
Links are nearly impossible impossible to discern from regular black text (at least on my monitor).

I don't really care about how specifically they're made more visible, but issues are disabled on the repo so I had to submit something!

I added back standard link underlines, but removed them in the nav where they're unneccessary.
I chose a random decent-looking blue color for links, which passes WCAG AA: https://webaim.org/resources/contrastchecker/?fcolor=0968DE&bcolor=FFFFFF

Before:
screen shot 2018-08-31 at 7 37 49 pm

After:
screen shot 2018-08-31 at 8 36 39 pm

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -81,11 +81,7 @@ h3.bullet::after {
     width: 5rem;
 }
 a {
-    color: #0B6D88;
-    text-decoration: none;
-}
-a:hover {
-    color: #0C637B;
+    color: #0968DE;
 }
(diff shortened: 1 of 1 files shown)


guozhangwang (approved):
Thanks @courajs . And I agree that `#0B6D88` and `#0C637B` are both not very contrast and they are very similar to each other as well.

PR #169, opened 2018-11-10
Update legacy doc paths

lindong28:


lindong28:
@guozhangwang The release process page https://cwiki.apache.org/confluence/display/KAFKA/Release+Process says that we should make this change for every release. The code in `includes/_header.htm` suggests that this change is needed to redirect the request related to the streams wiki.

Since this change has not been made in the past few releases, I am wondering if the change is actually needed. Could you review this PR? Thanks!

guozhangwang:
@lindong28 thanks for brining this up. @derrickdoo Could you double check and confirm if it is still needed today? If yes then why we did not discover any issues on the website?

derrickdoo:
@lindong28 @guozhangwang you don't need to update this with each release. This was just there to take care of old links prior to 0101.

lindong28:
@derrickdoo @guozhangwang Thanks for the information. I updated the release process with the change in https://cwiki.apache.org/confluence/pages/diffpagesbyversion.action?pageId=34840886&selectedPageVersions=143&selectedPageVersions=144 so that we do not need to update `_header.htm` for future release.

PR #171, opened 2018-11-21
Update documentation links to 2.1

omkreddy:


omkreddy:
@lindong28 @ijuma minor change to update doc links

lindong28:
@omkreddy Thanks for catching and fixing this issue! I updated the release process wiki (see https://cwiki.apache.org/confluence/pages/diffpagesbyversion.action?pageId=34840886&selectedPageVersions=144&selectedPageVersions=145) so that release manager can know what to update in the future.

Instead of listing all files that need to be updated, I updated the wiki to suggest that we can find those files by doing `grep -A1 "should always link the the latest" . -Irn`. Currently this string appears in almost all files that need to be updated. The only exceptions are `intro.html` and `quickstart.html`.

Can you also help put the string `<!-- should always link the the latest release's documentation -->` as appropriate in these two files?

omkreddy:
@lindong28 Updated the PR

lindong28:
Thanks for fixing this!

PR #172, opened 2018-11-22
Fix url for keys in download.html

lindong28:


lindong28:
According to https://www.apache.org/dev/release-signing.html#keys-policy, it says `...the Ant KEYS file is in the directory http://www.apache.org/dist/ant`. So it appears that we should use https://www.apache.org/dist/kafka/KEYS as key url in downloads.html.

PR #174, opened 2018-12-23
fix for broken zero-copy URL

russau:
The old developerWorks URL has been moved to a `developer.ibm.com` domain.


guozhangwang (approved):
Thanks @russau !

PR #176, opened 2019-01-06
KAFKA-7768: Add version to java html urls

ouertani:
small fix related to javadoc and version [KAFKA-7768](https://issues.apache.org/jira/browse/KAFKA-7768?filter=-2 )


guozhangwang:
@ouertani could you submit a PR for `kafka` under `docs` as well? Whenever a new version is released the content will be copy-pasted to `kafka-site` so if you do not change that repo, your fix here will still be overridden upon new release.

ouertani:
> @ouertani could you submit a PR for `kafka` under `docs` as well? Whenever a new version is released the content will be copy-pasted to `kafka-site` so if you do not change that repo, your fix here will still be overridden upon new release.
@guozhangwang 
--> https://github.com/apache/kafka/pull/6094

PR #178, opened 2019-01-07
KAFKA-7768: Use absolute paths for javadoc

vvcephei:
The breakage observed in KAFKA-7768 was only when following a link from a page with a non-versioned URL.

When starting from a versioned url, the number of ../ up-paths was correct, so inserting the version actually breaks it. To avoid this situation, this PR switches to using absolute paths in conjunction with the version number, as other links in the docs do.

See also #176, https://github.com/apache/kafka/pull/6094, and https://github.com/apache/kafka/pull/6100


vvcephei:
Hi @guozhangwang , do you mind also taking a look at this?

Thanks,
-John

PR #182, opened 2019-01-23
MINOR: Use lists.apache.org for mailing list archives

mimaison:
lists.apache.org has a much better UI for browsing than mail-archives.apache.org


guozhangwang (approved):
Tested all three urls, works on my browser.

Thanks @mimaison .

PR #188, opened 2019-02-23
MINOR: Add PGP key for bbejeck

bbejeck:
Adding key


KEYS
```diff
@@ -1070,3 +1070,63 @@ gm+FU3WdD9oqC3/dDKZ7pFizUhijUZo+nfedhj+7+ZGPFrPX11TS+N3p2E863DGG
 gGDok4uiZw4HdrV8d3/JvahBXfQWH12KeoFmZi/pY/XiPZszEyPrdRVrNbRag/0=
 =V7qY
 -----END PGP PUBLIC KEY BLOCK-----
+pub   rsa4096 2019-02-23 [SC]
+      1BE3 316E 762D 17FF 4D2D  55B6 7D19 0980 87E7 3B62
+uid           [ultimate] Bill Bejeck (CODE SIGNING KEY) <bbejeck@gmail.com>
```
mjsax, line 1075: I think, your key should (must?) be for your `@apache.org` email address

bbejeck: Well, the docs say it's recommended (not required) I guess I overlooked that part. I could update and re-push to a key-server but as I looked through the `KEYS` file, and it seems several of them don't use the `@apache.org` address.  If you insist I'll update the uid.

mjsax: I was not sure. It's up to you. I personally don't care. Just wanted to make sure, it does not slip if it would have been a requirement.

mjsax (approved):

bbejeck:
Merged #188

PR #189, opened 2019-03-09
MINOR: fix Streams config

mjsax:
This was fixed in kafka.git, but we never updated the web docs (those are only updated when we do a new release).


mjsax:
Call for review @guozhangwang @bbejeck @vvcephei @ableegoldman

bbejeck (approved):
LGTM

PR #233, opened 2019-10-25
Update site for 2.3.1 release

mumrah:


hachikuji (approved):
LGTM

index.html
```diff
@@ -21,6 +21,10 @@ <h2>Store</h2>
 			</a>
 			<div class="news">
 				<h2>Latest News</h2>
+				<article>
+					<h3><a href="https://kafka.apache.org/downloads#2.3.1">AK Release 2.3.1</a></h3>
+					<time>October 24, 2019</time>
+				</article>
```
mjsax, line 27: @mumrah We should remove the entry about `2.3.0` release

downloads.html
```diff
@@ -11,6 +11,32 @@ <h1>Download</h1>
     You can verify your download by following these <a href="https://www.apache.org/info/verification.html">procedures</a> and using these <a href="https://www.apache.org/dist/kafka/KEYS">KEYS</a>.
     </p>
 
+    <span id="2.3.1"></span>
```
mjsax, line 14: Line 8 above also need to be updated

index.html
mumrah: I wasn't sure about that since these `<article>` seem pretty new. I'll remove the 2.3.0 section

downloads.html
mumrah: Fixed.

index.html
mjsax: It is quite new, but why "advertise" 2.3.1 and 2.3.0? Thanks for the follow up PR.

PR #236, opened 2019-11-10
MINOR fix description about whether internal topics respect broker defaults

alfredxiao:
Previous description says `Kafka Streams automatically creates internal repartitioning and changelog topics, and does not use broker auto-topic creation. Therefore, internal topics take on the default broker configurations for new topics.`
This is incorrect and self-conflicting. Because internal topics actually do not respect broker defaults for newly created topics.

### DIFF
--- a/10/streams/developer-guide/manage-topics.html
+++ b/10/streams/developer-guide/manage-topics.html
@@ -84,8 +84,10 @@
       <p>
         Kafka Streams automatically creates internal repartitioning and changelog
         topics, and does <em>not</em> use broker auto-topic creation.
-        Therefore, internal topics do not take on the default broker configurations for
-        new topics.
+        Therefore, internal topics do not take on all the default broker configurations for
+        new topics. In particular, <code>replication.factor</code> (which defaults to <code>1</code>),
+        <code>num.partitions</code>, and <code>cleanup.policy</code> do not respect broker configurations
+        fot automatically created topics.
         You can override the default configs used when creating these topics by
         adding any configs from <code>TopicConfig</code> to your
         <code>StreamsConfig</code> with the prefix
--- a/10/streams/developer-guide/manage-topics.html
+++ b/10/streams/developer-guide/manage-topics.html
@@ -84,7 +84,7 @@
       <p>
         Kafka Streams automatically creates internal repartitioning and changelog
         topics, and does <em>not</em> use broker auto-topic creation.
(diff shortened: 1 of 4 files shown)


10/streams/developer-guide/manage-topics.html
```diff
@@ -84,7 +84,7 @@
       <p>
         Kafka Streams automatically creates internal repartitioning and changelog
         topics, and does <em>not</em> use broker auto-topic creation.
-        Therefore, internal topics take on the default broker configurations for
+        Therefore, internal topics do not take on the default broker configurations for
```
mjsax: Thinking about this once again, some broker defaults are actually used... Just not all. Maby best to rephrase as `do not take on [all] the default`?

Might be worth to clarify which configs? In particular number of partitions and replication factor, as well as, cleanup strategy (and others -- depending on the version, but maybe calling out those three is sufficient?)

alfredxiao: Good thought to be more accurate. I'll try to digest what configs are not respecting broker defaults, which could be changed between versions. Otherwise, might just follow your suggestion. Thanks

alfredxiao: Couldn't find details more accurate than those you mentioned. I am happy to apply what you suggested. @mjsax

mjsax:
Thanks for the PR @alfredxiao -- LGTM. This PR will update the webpage directly. However, the source of truth is the main Kafka repository. Can you do a PR against `trunk` with the same change? (I can cherry pick that PR to older branches -- no need to open PRs against all branches).

alfredxiao:
@mjsax Think you were referring to `trunk` of the `https://github.com/apache/kafka` codebase.
However, it appears that the very above paragraph/description does not exist in the section about `Internal Topics` in file `https://github.com/apache/kafka/blob/trunk/docs/streams/developer-guide/manage-topics.html`, including branches from 1.x, 2.x to trunk.

mjsax:
It seems that the whole paragraph is missing in `trunk` -- I guess somebody did a PR against this repro with a PR against `trunk`... _sigh_

Can you do a PR against `trunk` that adds the full paragraph? This PR should also update newer versions `22` and `23` (if the paragraph is missing, also add the full paragraph to get both repos in-sync).

Thanks a lot!

mjsax:
Closing this PR as abandoned. (We can reopen if necessary.)

PR #247, opened 2019-12-28
MINOR: fix omitted 'next.' in interactive queries documentation

dongjinleekr:
```java
while (range.hasNext()) {
  KeyValue<String, Long> next = range.next();
  System.out.println("count for " + next.key + ": " + value);
}
```

(from: [Interactive Queries Documentation (2.4)](https://kafka.apache.org/24/documentation/streams/developer-guide/interactive-queries.html))

`value` should be `next.value`.


guozhangwang:
Thanks for the find @dongjinleekr , should we check if other versions than 0110 also need to be fixed? Also in `kafka` code repo if you could look into the `docs` folder and see if that also need to be fixed to via another PR would be great.

dongjinleekr:
@guozhangwang Sure. Let me see. :smile:

dongjinleekr:
@guozhangwang Here is the check; it seems like there are no other same typos. Also, I opened a PR for the counterpart. :smiley:

kafka-docs

guozhangwang:
I think they are still there in the newer branches, but just because they are reformatted (in a bad way) it is hard to find. For example in 1.0:

https://github.com/apache/kafka-site/blob/asf-site/10/streams/developer-guide/interactive-queries.html#L161

See in https://kafka.apache.org/24/documentation/streams/developer-guide/interactive-queries.html#querying-local-key-value-stores the formatted code example is also incorrect.

dongjinleekr:
@guozhangwang Here it is; I found the glitches in the docs of version 22, 23, and 24. Since they are the only still supported, I reverted version 0110 instead.

Before: `<span class="n">value</span><span class="o">);</span>`
After: `<span class="n">next</span><span class="o">.</span><span class="na">value</span><span class="o">);</span>`

(see: apache/kafka#7883)

guozhangwang:
Thanks @dongjinleekr !

PR #250, opened 2020-01-14
Add CVE-2019-12399

rhauch:


cve-list.html
```diff
@@ -8,6 +8,40 @@ <h1>Apache Kafka Security Vulnerabilities</h1>
 
 This page lists all security vulnerabilities fixed in released versions of Apache Kafka.
 
+<h2><a href="http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-12399">CVE-2019-12399</a>
+Apache Kafka Connect REST API may expose plaintext secrets in tasks endpoint</h2>
+
+<p>When Connect workers in Apache Kafka 2.0.0, 2.0.1, 2.1.0, 2.1.1, 2.2.0, 2.2.1, or 2.3.0 are
```
cmccabe, line 14: It would be nice to split this into more than one sentence, since it's quite long

cmccabe:
LGTM after fixing minor style nit

cve-list.html
rhauch: It is long, but the vulnerability only can occur when all of those criteria are met. Not sure how to break up the sentences without making it seem less restricted.

rhauch: Also, for better or worse, it's the same text that's in the CVE.

cmccabe: It could be broken up to be more readable.  If it's duplicating the CVE, though, then let's keep it as-is.

PR #259, opened 2020-04-14
Add upgrade notes for 2.5.0 release

mumrah:


omkreddy (approved):
@mumrah Thanks for the PR. LGTM.
we also need to update the kafka repo docs on trunk and 2.5 branches.

PR #261, opened 2020-04-16
MINOR: update download links to point to ASF archive servers

mjsax:
Call for review @mumrah 

There is 2.4.1 and 2.3.1, hence, 2.4.0 and 2.3.1 should point to the archive.


mumrah (approved):
Verified the links, looks good to me 👍. The PR description should read `2.3.0` instead of `2.31`

PR #262, opened 2020-04-17
Switch order of sections on tumbling and hopping windows in streams doc

soenkeliebau:
Equivalent to cc4e3aa30207679f70f41abd2ca2c79a5d91ca9d on apache/kafka repo.

Switch order of sections on tumbling and hopping windows in streams doc. Tumbling windows are defined as "special case of hopping time windows" - but hopping windows currently only explained later in the docs. (#8505)

Currently, tumbling windows are defined as "a special case of hopping time windows" in the streams docs, but hopping windows are only explained in a subsequent section.
I think it would make sense to switch the order of these paragraphs around. To me this also makes more sense semantically.

Testing
Built the site and checked that everything looks ok and html is valid (or at least didn't contain any new warnings that were caused by this change).

### DIFF
--- a/25/streams/developer-guide/dsl-api.html
+++ b/25/streams/developer-guide/dsl-api.html
@@ -55,8 +55,8 @@
                         </ul>
                         </li>
                         <li><a class="reference internal" href="#windowing" id="id19">Windowing</a><ul>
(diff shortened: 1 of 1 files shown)


bbejeck (approved):
Thanks @soenkeliebau LGTM.

Validated by deploying locally.

bbejeck:
Thanks again @soenkeliebau for the docs improvement!

PR #266, opened 2020-05-27
MINOR: Add more info in the "adding to contributor list" section

showuon:
We usually got this kind of email to request to be added into contributor list, but without JIRA username provided. That wasted the committer's or the existing contributor's time to do more explanation.  I found it's because we didn't provide enough information in the [How to contribute](https://kafka.apache.org/contributing) page. I added the info explicitly: **with your JIRA account username provided in the email** .

**example 1**
image

**example 2**
image


showuon:
hi @mjsax , I think you must love this PR because you won't have to handle the "requesting to be added into contributor list" but without JIRA ID provided issue. :)

mjsax:
Thanks a lot @showuon! This seems to be super helpful!

PR #272, opened 2020-07-01
KAFKA-10222:Incorrect methods show up in 0.10 Kafka Streams docs

huxihx:
https://issues.apache.org/jira/browse/KAFKA-10222

Non-existent methods show up in the doc:
_builder.from("my-input-topic").mapValue(value -> value.length().toString()).to("my-output-topic");_

There is no method named `from` or `mapValues`. They should be `stream` and `mapValues` respectively.


huxihx:
@ijuma Please review this patch. Thanks.

mjsax:
Thanks for the PR @huxihx -- note, that this pages are generated from the JavaDocs. Thus, if we fix it, we should fix it in the source code?

huxihx:
@mjsax Thanks for reminding. Will close this PR and open a new PR to track this issue.

mjsax:
We can still merge this PR, because otherwise the web-page would not be updated.

PR #282, opened 2020-08-05
bug fixes

scott-confluent:
This is a first round fix for the bugs we've found. This is for all high priority bugs unless otherwise noted in the google sheet we're tracking them with

### home page
-  spelling
- re-introduce see full list tooltip

### docs
- remove version banner
- fixed Markup issues on https://kafka.apache.org/documentation/streams/developer-guide/dsl-api.html
- removed duplicate headline and initial section on streams running app page

### powered by
- fixed JSON issue in page

### DIFF
--- a/25/documentation.html
+++ b/25/documentation.html
@@ -31,7 +31,7 @@
     <!--#include virtual="toc.html" -->
   </div>
 	<div class="right">
-    <!--#include virtual="../includes/_docs_banner.htm" -->
+    <!--//#include virtual="../includes/_docs_banner.htm" -->
     <a href="#top" id="docs-top-link">&#10224;<br />Top</a>
     <h1 class="docs-title">Documentation</h1>
     <h3>Kafka 2.5 Documentation</h3>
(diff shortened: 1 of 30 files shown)


25/streams/developer-guide/write-streams.html
```diff
@@ -90,22 +90,22 @@
...
-    <artifactId>kafka-streams</artifactId>
-    <version>{{fullDotVersion}}</version>
-</dependency>
-<dependency>
-    <groupId>org.apache.kafka</groupId>
-    <artifactId>kafka-clients</artifactId>
-    <version>{{fullDotVersion}}</version>
-</dependency>
+          <pre class="line-numbers"><code class="language-xml">&lt;dependency>
+    &lt;groupId>org.apache.kafka&lt;/groupId>
```
guozhangwang, line 94: Should the `>` be replaced too?

25/documentation.html
```diff
@@ -31,7 +31,7 @@
     <!--#include virtual="toc.html" -->
   </div>
 	<div class="right">
-    <!--#include virtual="../includes/_docs_banner.htm" -->
```
guozhangwang, line 34: Ditto for other 25/xyx.html pages: does the same issue exist in older versions (24, 23...)?

guozhangwang (commented):
made a pass , cc @ijuma @bbejeck for another pass.

25/streams/developer-guide/write-streams.html
scott-confluent: theoretically it should be generated with html entities for both gt and lt, but the opening is the only one that matters for highlighting the syntax. it's not an easy thing to find/replace so we'd have to do an audit of all the docs.

25/documentation.html
scott-confluent: I believe the older versions SHOULD have this banner.

PR #284, opened 2020-08-06
MINOR: Change the arrow direction based on the view state is expanded or not

showuon:
Change the arrow direction based on the view state is expanded or not. We can identify it's in expanded mode or not by checking the return value of toggle method. Reference: https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle

image

image

**after my fix when expanded**
image

### DIFF
--- a/includes/_footer.htm
+++ b/includes/_footer.htm
@@ -329,7 +329,13 @@
 				var docsHandle = document.querySelector('.toc-handle');
 
 				function toggleDocsWidth() {
-					docsContainer.classList.toggle('expanded');
+					let isExpanded = docsContainer.classList.toggle('expanded');
+					// change the arrow direction based on the view state is expanded or not
+					if (isExpanded) {
+						docsHandle.textContent = ">";
+					} else {
+						docsHandle.textContent = "<";
+					}
 				}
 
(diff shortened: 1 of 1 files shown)


showuon:
@scott-confluent @guozhangwang , could you help review this PR? Thanks.

includes/_footer.htm
```diff
@@ -329,7 +329,13 @@
 				var docsHandle = document.querySelector('.toc-handle');
 
 				function toggleDocsWidth() {
-					docsContainer.classList.toggle('expanded');
+					let isExpanded = docsContainer.classList.toggle('expanded');
+					// change the arrow direction based on the view state is expanded or not
+					if (isExpanded) {
```
scott-confluent, line 334: This is technically only working because JavaScript is lax on types and the `DOMTokenList` returned by  `classList.toggle` is empty resulting in a falsy value. We should probably do `isExapnded.contains('expanded')` as the condition to be more explicit.

scott-confluent (approved):

showuon:
@guozhangwang , I think this PR is good to merge except you have other opinions. Thanks.

guozhangwang (approved):
Thanks @showuon

PR #285, opened 2020-08-07
MINOR: Fix the wrong doc path and missing zk/docker button in quickstart page

showuon:
1. In quickstart page, we should see the zookeeper and docker button for user to select, just like the `quickstart-docker` page showed: https://kafka.apache.org/quickstart-docker

image

2. In quickstart page, we'll see the following error message displayed. After investigation, I found it's because we point to the wrong document `26/quickstart-zookeeper.html`. Change to the correct name `25/quickstart-zookeeper.html` to fix it.

image

### DIFF
--- a/quickstart.html
+++ b/quickstart.html
@@ -16,10 +16,18 @@
           allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowfullscreen>
         </iframe>
       </div>
-      
+
+      <ul class="page-header-nav">
+        <li class="page-header-nav-item">
+          <a href="/quickstart" class="page-header-nav-item-anchor current">Zookeeper</a>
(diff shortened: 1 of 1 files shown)


showuon:
@scott-confluent @rhauch , could you review this PR ? Thanks.

rhauch:
Thanks for identifying the problem and proposing a fix, @showuon. However, I don't think referencing the quickstarts in `25` directory is what we want -- instead, we want to add the quickstarts to the `26` directory.

I've created #286 to address the issue more consistently with the existing patterns.

rhauch:
Close without merging.

rhauch:
https://kafka.apache.org/quickstart appears to be working again after merging #286.

PR #287, opened 2020-08-10
fixes for intro page and various docs page headings

scott-confluent:


intro.html
```diff
@@ -20,7 +20,203 @@ <h1 class="page-header-title">Introduction</h1>
       
     </div>
 <!-- should always link the the latest release's documentation -->
-    <!--#include virtual="/26/introduction.html" -->
+    <h4 class="anchor-heading">
```
guozhangwang: Why we have to copy-paste now? The old way is more convienent since we do not need to copy-paste everytime we update the newest released doc.

PR #286, opened 2020-08-10
MINOR: Fix renamed and reformatted quickstart, broken since 2.6.0 release

rhauch:
The AK code contains a `quickstart.html` file, so that when 2.6.0 was released the newer `quickstart-docker.html` and `quickstart-zookeeper.html` were not included in the `26` directory, breaking the https://kafka.apache.org/quickstart page.

This recovers those files into the `asf-site` branch's `26` directory.

Supersedes #285

PR #289, opened 2020-08-11
MINOR: Fix the broken past-events link

showuon:
Fix the broken past-events link. It should be in the `kafka-summit.org` domain, not under `kafka.apache.org`.

### DIFF
--- a/25/quickstart-zookeeper.html
+++ b/25/quickstart-zookeeper.html
@@ -267,7 +267,7 @@ wordCounts.toStream().to("output-topic"), Produced.with(Serdes.String(), Serdes.
         -->
         <li>
             Join a <a href="/events">local Kafka meetup group</a> and
-            <a href="/past-events">watch talks from Kafka Summit</a>,
+            <a href="https://kafka-summit.org/past-events/">watch talks from Kafka Summit</a>,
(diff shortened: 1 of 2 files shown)


showuon:
@scott-confluent @mjsax , could you help review this PR? Thanks.

guozhangwang (commented):
It seems

```
https://kafka.apache.org/26/documentation/#quickstart
```

and 

```
https://kafka.apache.org/25/documentation/#quickstart
```

Are broken with `[an error occurred while processing this directive]` (and also `https://kafka.apache.org/documentation/#quickstart` as it is pointing to 26 now) while older branches are fine.

I think we should fix these as well. wdyt @scott-confluent

showuon:
It's because the `quickstart.html` got removed in this PR: this PR: https://github.com/apache/kafka-site/pull/286. I think we should keep `quickstart.html`, `quickstart-zookeeper.html` and `quickstart-docker.html` files in the `25` and `26` folder. What do you think @scott-confluent ?

showuon:
@guozhangwang , the issue you mentioned is already addressed in this PR: https://github.com/apache/kafka-site/pull/293. Not sure if you have other comments for this PR? Thanks.

PR #294, opened 2020-08-12
MINOR: removing docs version header from 2.6 docs

scott-confluent:
Also re-adding it to the 2.5 docs

### DIFF
--- a/25/streams/architecture.html
+++ b/25/streams/architecture.html
@@ -174,7 +174,7 @@
 <div class="content documentation ">
     <!--#include virtual="../../includes/_nav.htm" -->
     <div class="right">
-        <!--//#include virtual="../../includes/_docs_banner.htm" -->
+        <!--#include virtual="../../includes/_docs_banner.htm" -->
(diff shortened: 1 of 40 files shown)


ableegoldman:
Should this header be on all the earlier versions too? I don't see it anywhere at the moment (except for 2.6)

rhauch:
@ableegoldman I see the header on 2.4 and earlier. IIUC, this is to *add* this header to the 2.5 docs and *remove* this header from all of the 2.6 docs (e.g., https://kafka.apache.org/26/documentation/streams/developer-guide/).

rhauch (approved):
LGTM. Thanks, @scott-confluent!

ableegoldman:
Huh, I don't see the header on 2.4 (or earlier). Since this has been merged I now only see it on 2.5

ableegoldman:
Just tried in a different browser and I still only see the header on 2.5. The 2.4 formatting looks a little different, but no "You're viewing an older version..." header

PR #296, opened 2020-08-18
MINOR: fix for extra line break in streams processor API documentation

SanjayYr:
Removed the extra "/p" tag.


mjsax:
@SanjayYr Does this need to be fixed in `kafka.git`, too, and also for newer version?

mjsax:
Seems this part was re-written via https://github.com/apache/kafka/commit/a08e0cfe652dc93cbed07a8912a460d85c579dd8

No need to update `kafka.git` for old versions.

Will do a follow up fix for other versions.

PR #297, opened 2020-08-18
MINOR: fix for example implementation of how to add a new header to the record

SanjayYr:
Fixed method invocation and arguments.


mjsax:
@SanjayYr Does this need to be fixed in `kafka.git` and other versions, too?

mjsax:
Seems it's fixed in 3.0+

Will do a follow up PR for other 2.0 versions.

PR #298, opened 2020-08-19
MINOR: remove unreleased versions from CVE page

mjsax:
As reported on the mailing list: https://lists.apache.org/list.html?dev@kafka.apache.org:lte=1M:cve%202.0.2%202.1.2

Call for review @ijuma


ijuma (approved):
LGTM

PR #301, opened 2020-09-04
Redo 2.6 docs commit

ableegoldman:
Just noticed that all of the docs I added just before 2.6 aren't actually showing up on the site. Seems like they were accidentally reverted by [this commit](https://github.com/apache/kafka-site/commit/35d3b804b86c249aace9a729c5abd9be4525ef3d)

So, this PR just redoes everything that was originally in https://github.com/apache/kafka-site/commit/4c19eab4f6315c91549be0bff08ad46173be4f64

### DIFF
--- a/26/streams/developer-guide/memory-mgmt.html
+++ b/26/streams/developer-guide/memory-mgmt.html
@@ -202,7 +202,10 @@
        <span class="o">}</span>
     <span class="o">}</span>
       </div>
-        <sup id="fn1">1. INDEX_FILTER_BLOCK_RATIO can be used to set a fraction of the block cache to set aside for "high priority" (aka index and filter) blocks, preventing them from being evicted by data blocks. See the full signature of the <a class="reference external" href="https://github.com/facebook/rocksdb/blob/master/java/src/main/java/org/rocksdb/LRUCache.java#L72">LRUCache constructor</a>. </sup>
+        <sup id="fn1">1. INDEX_FILTER_BLOCK_RATIO can be used to set a fraction of the block cache to set aside for "high priority" (aka index and filter) blocks, preventing them from being evicted by data blocks. See the full signature of the <a class="reference external" href="https://github.com/facebook/rocksdb/blob/master/java/src/main/java/org/rocksdb/LRUCache.java#L72">LRUCache constructor</a>.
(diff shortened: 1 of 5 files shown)


26/streams/developer-guide/memory-mgmt.html
```diff
@@ -202,7 +202,10 @@ <h2><a class="toc-backref" href="#id3">RocksDB</a><a class="headerlink" href="#r
...
+        <sup id="fn1">1. INDEX_FILTER_BLOCK_RATIO can be used to set a fraction of the block cache to set aside for "high priority" (aka index and filter) blocks, preventing them from being evicted by data blocks. See the full signature of the <a class="reference external" href="https://github.com/facebook/rocksdb/blob/master/java/src/main/java/org/rocksdb/LRUCache.java#L72">LRUCache constructor</a>.
+          NOTE: the boolean parameter in the cache constructor lets you control whether the cache should enforce a strict memory limit by failing the read or iteration in the rare cases where it might go larger than its capacity. Due to a
+          <a class="reference external" href="https://github.com/facebook/rocksdb/issues/6247">bug in RocksDB</a>, this option cannot be used
+          if the write buffer memory is also counted against the cache. If you set this to true, you should NOT pass the cache in to the <code>WriteBufferManager</code> and just control the write buffer and cache memory separately.</sup>
```
vvcephei, line 208: It looks like this (and some other stuff in the config-streams file) has come along for the ride. We should make sure it's referencing behavior that was released, not behavior that will be released next. Can you double-check that this actually is in reference to 2.6.0, and not 2.6.1+?

vvcephei (approved):
Thanks, @ableegoldman , and sorry the original commit got smashed.

Can you follow up on the question I asked below?
Also, can you confirm that you rendered the docs locally and verified the formatting of the files you're changing?

Other than those concerns, this LGTM!

26/streams/developer-guide/memory-mgmt.html
ableegoldman: Nope, this is correct. My original commit had a few other fixes on the side in addition to the KIP-441/KIP-613 docs

As far as I know this has always been the case in every RocksDB version we've depended on but didn't discover the bug until just recently. (I believe it actually is fixed in some newer versions, so we may be able to remove this again once we upgrade)

PR #303, opened 2020-10-16
Add dajac to committers

dajac:


junrao (approved):
@dajac : LGTM. Congratulation!

PR #306, opened 2020-10-28
Add ableegoldman to committers

ableegoldman:
A bit late on this but here we go 🙂


committers.html
```diff
@@ -334,6 +334,14 @@ <h1 class="content-title">The committers</h1>
           Committer<br>
           <a href="https://www.linkedin.com/in/chia7712/">/in/chia7712</a><br>
         </td>
+        <td>
```
chia7712, line 339: It needs a new line (```<tr>``` and ```</tr>```)

ableegoldman: Ah, good catch. Fixed now -- thanks!

chia7712 (approved):

PR #307, opened 2020-11-09
[MINOR] fixing docs banner so it appears correctly on older versions

scott-confluent:


bbejeck:
~\cc @mimaison these changes need to go into 2.6.1 after 2.7 is released~

scott-confluent:
## NOTE: this should NOT be merged until 2.7 is officially released

mjsax:
We also don't have the banner in `2.5` atm -- can we do a parallel PR to get it back in `2.5` right away?

mimaison:
@bbejeck I'm not sure I understand why this is blocking 2.6.1.

Can I port these changes back into kafka/2.6 and start the release process? As far as I can tell, we'll just need to update the website for 2.6.1 after 2.7.0 is released. But we can get started on RCs and votes. And even if a vote passes before 2.7.0 is out, just wait a bit before completing the release.

bbejeck:
@mimaison this shouldn't block 2.6.1; my apologies; I misunderstood what needed to happen.

After 2.7 get released, I'll merge this PR, but for now, there's nothing you need to do as 2.6.x is the current version.
\cc @scott-confluent

bbejeck (approved):

PR #311, opened 2020-11-11
MINOR: Fix the wrong HTML format in powered by page

showuon:
When visiting the powered by page, I suddenly found the hyperlink text is weird in SemaText.
image

After investigation, I found it's because we didn't end the `<a>` tag content well. We missed some right angle bracket (>) and some single quote ('), so that some text are disappeared on the page. Ex:
image

We should end with the text `here` with hyperlink.

**Fix those issues!**
example after fixed:
image

### DIFF
--- a/powered-by.html
+++ b/powered-by.html
@@ -91,7 +91,7 @@
         "link": "http://www.cerner.com/",
         "logo": "cerner.png",
         "logoBgColor": "#ffffff",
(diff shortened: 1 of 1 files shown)


showuon:
@guozhangwang @scott-confluent , please help review. Thanks.

showuon:
@guozhangwang @scott-confluent , please help review. Thanks.

guozhangwang:
I did a quick check and can confirm all the pointed out issues are valid on site. Great job @showuon ! Merging now.

PR #324, opened 2021-01-22
KAFKA-8930: MirrorMaker v2 documentation

miguno:
This adds a new user-facing documentation "Geo-replication (Cross-Cluster Data Mirroring)" section to the Kafka Operations documentation that covers MirrorMaker v2.

### DIFF
--- a/27/ops.html
+++ b/27/ops.html
@@ -92,25 +92,11 @@
   <p></p>
   However if racks are assigned different numbers of brokers, the assignment of replicas will not be even. Racks with fewer brokers will get more replicas, meaning they will use more storage and put more resources into replication. Hence it is sensible to configure an equal number of brokers per rack.
 
-  <h4 class="anchor-heading"><a id="basic_ops_mirror_maker" class="anchor-link"></a><a href="#basic_ops_mirror_maker">Mirroring data between clusters</a></h4>
+  <h4 class="anchor-heading"><a id="basic_ops_mirror_maker" class="anchor-link"></a><a href="#basic_ops_mirror_maker">Mirroring data between clusters & Geo-replication</a></h4>
 
-  We refer to the process of replicating data <i>between</i> Kafka clusters "mirroring" to avoid confusion with the replication that happens amongst the nodes in a single cluster. Kafka comes with a tool for mirroring data between Kafka clusters. The tool consumes from a source cluster and produces to a destination cluster.
-
-  A common use case for this kind of mirroring is to provide a replica in another datacenter. This scenario will be discussed in more detail in the next section.
   <p>
-  You can run many such mirroring processes to increase throughput and for fault-tolerance (if one process dies, the others will take overs the additional load).
-  <p>
-  Data will be read from topics in the source cluster and written to a topic with the same name in the destination cluster. In fact the mirror maker is little more than a Kafka consumer and producer hooked together.
-  <p>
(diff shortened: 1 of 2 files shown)


27/ops.html
```diff
@@ -553,7 +539,558 @@ <h3 class="anchor-heading"><a id="datacenters" class="anchor-link"></a><a href="
...
+  </ul>
+
+  <p>
+  <em>Note: Geo-replication with MirrorMaker replicates data across Kafka clusters. This inter-cluster replication is different from Kafka's <a href="#replication">intra-cluster replication</a>, which replicates data within the same Kafka cluster.</em>
+  </p>
+
+  <h4 class="anchor-heading"><a id="georeplication-flows" class="anchor-link"></a><a href="#georeplication-flows">What Are Replication Flows</a></h4>
+
+  <p>
+    With MirrorMaker, Kafka administrators can replicate topics, topic configurations, consumer groups and their offsets, and ACLs from one or more source Kafka clusters to one or more target Kafka clusters, i.e., across cluster environments. In a nutshell, MirrorMaker consumes data from the source cluster with source connectors, and then replicates the data by producing to the target cluster with sink connectors.
```
ryannedolan: "with sink connectors" is not true at the moment, since I don't think we have a sink connector yet. And even when we do, it would usually be sufficient to use source _or_ sink connector. There are certainly cases where this sentence is true, but I think it's misleading as a general statement.

Maybe "In a nutshell, MirrorMaker uses Connectors to consume from source clusters and produce to target clusters" or something like that.

27/ops.html
```diff
@@ -553,7 +539,558 @@ <h3 class="anchor-heading"><a id="datacenters" class="anchor-link"></a><a href="
...
+us-east.ssl.truststore.password=my-secret-password
+us-east.ssl.keystore.location=/path/to/keystore.jks
+us-east.ssl.keystore.password=my-secret-password
+us-east.ssl.key.password=my-secret-password
+</code></pre>
+
+  <h5 class="anchor-heading"><a id="georeplication-topic-naming" class="anchor-link"></a><a href="#georeplication-topic-naming">Custom Naming of Replicated Topics in Target Clusters</a></h5>
+
+  <p>
+    Replicated topics in a target cluster—sometimes called <em>remote</em> topics—are renamed according to a replication policy. MirrorMaker uses this policy to ensure that events (aka records, messages) from different clusters are not written to the same topic-partition. By default as per <a href="https://github.com/apache/kafka/blob/trunk/connect/mirror-client/src/main/java/org/apache/kafka/connect/mirror/DefaultReplicationPolicy.java">DefaultReplicationPolicy</a>, the names of replicated topics in the target clusters have the format <code>{source}.{source_topic_name}</code>:
```
ryannedolan, line 828: I think "records" is more prevalent in Kafka docs vs "events". Maybe verify that and stick with whatever the rest of the docs use.

ryannedolan (commented):
This is great! Couple small comments so far.

27/ops.html
miguno: The use of "events" has become more prevalent, see e.g. the updated introduction page (https://kafka.apache.org/intro) and quickstart (https://kafka.apache.org/quickstart). Hence I prefer to leave the text as-is. The included "(aka records, messages)" parenthesis makes the nomenclature sufficiently clear to the reader, imho.

27/ops.html
miguno: Thanks, @ryannedolan. Text updated.

bbejeck (approved):
LGTM!

I validated by rendering the changes locally

omkreddy:
We should also add these docs to `kafka/docs` repo

bbejeck:
> We should also add these docs to kafka/docs repo

@omkreddy, yes a PR for kafka/docs is coming soon

miguno:
kafka/docs PR is up at https://github.com/apache/kafka/pull/9983

PR #327, opened 2021-01-28
Use downloads.apache.org for current releases

mimaison:
As per https://www.apache.org/legal/release-policy.html, projects should use downloads.apache.org for current release artifacts instead of www.apache.org/dist.


mimaison:
@mjsax Can you take a look? Thanks

mjsax (approved):
Nice find! Seems to be a new policy. The old links get redirected to the new location anyway. Good to update it.

PR #330, opened 2021-02-08
MINOR: remove the indent in security doc

showuon:
Fix the kafka-site: this is the same fix as the kafka PR: https://github.com/apache/kafka/pull/10002.

@omkreddy , please help review. there's only 233 line changes here instead of 234 as the original PR, is because I accidentally remove the intent for a line. Fix in another PR: https://github.com/apache/kafka/pull/10081. Thanks.


omkreddy (approved):
LGTM

PR #331, opened 2021-02-19
MINOR: update the old anchor #intro_topic into the new one

showuon:
Accidentally found the anchor # intro_topic doesn't exist anymore, and cause the quick start page has a wrong link. Correct it as the new # intro_concepts_and_terms anchor

image


mimaison (approved):
LGTM. Thanks for the PR

PR #332, opened 2021-02-20
KAFKA-12350: correct the wrong default value in doc

showuon:
Correct the wrong default value. ref: https://github.com/apache/kafka/pull/10165


showuon:
@chia7712 , could you also review this PR? Same as this one: apache/kafka#10165, Just in kafka-site repo. Thanks.

chia7712:
@showuon Thanks for this patch :)

PR #333, opened 2021-02-24
Added documentation fix for KAFKA-12360

nicodds:
The fix clarify the use of max.task.idle.ms, as pointed out in https://issues.apache.org/jira/browse/KAFKA-12360


mjsax:
@nicodds Thanks for the PR!

We also need a second PR for https://github.com/apache/kafka that is the source of truth for the docs.

nicodds:
ehy @mjsax, thank you for you reply. I've made a PR also to the source code, here it is: https://github.com/apache/kafka/pull/10207

mjsax:
Closing due to inactivity. The corresponding ticket was resolved in the meantime.

PR #334, opened 2021-03-01
KAFKA-12393: Document multi-tenancy considerations

miguno:
See https://issues.apache.org/jira/browse/KAFKA-12393

### DIFF
--- a/27/ops.html
+++ b/27/ops.html
@@ -1090,7 +1090,157 @@ checkpoint-latency-ms-avg
   </p>
 
 
-  <h3 class="anchor-heading"><a id="config" class="anchor-link"></a><a href="#config">6.4 Kafka Configuration</a></h3>
+  <h3 class="anchor-heading"><a id="multitenancy" class="anchor-link"></a><a href="#multitenancy">6.4 Multi-Tenancy</a></h3>
+
+  <h4 class="anchor-heading"><a id="multitenancy-overview" class="anchor-link"></a><a href="#multitenancy-overview">Multi-Tenancy Overview</a></h4>
+
+  <p>
+    As a highly scalable event streaming platform, Kafka is used by many users as their central nervous system, connecting in real-time a wide range of different systems and applications from various teams and lines of businesses. Such multi-tenant cluster environments command proper control and management to ensure the peaceful coexistence of these different needs. This section highlights features and best practices to set up such shared environments, which should help you operate clusters that meet SLAs/OLAs and that minimize potential collateral damage caused by "noisy neighbors".
+  </p>
+
+  <p>
+    Multi-tenancy is a many-sided subject, including but not limited to:
+  </p>
+
+  <ul>
+    <li>Creating user spaces for tenants (sometimes called namespaces)</li>
+    <li>Configuring topics with data retention policies and more</li>
+    <li>Securing topics and clusters with encryption, authentication, and authorization</li>
+    <li>Isolating tenants with quotas and rate limits</li>
+    <li>Monitoring and metering</li>
+    <li>Inter-cluster data sharing (cf. geo-replication)</li>
+  </ul>
+
+  <h4 class="anchor-heading"><a id="multitenancy-topic-naming" class="anchor-link"></a><a href="#multitenancy-topic-naming">Creating User Spaces (Namespaces) For Tenants With Topic Naming</a></h4>
+
+  <p>
+    Kafka administrators operating a multi-tenant cluster typically need to define user spaces for each tenant. For the purpose of this section, "user spaces" are a collection of topics, which are grouped together under the management of a single entity or user.
+  </p>
+
+  <p>
+    In Kafka, the main unit of data is the topic. Users can create and name each topic. They can also delete them, but it is not possible to rename a topic directly. Instead, to rename a topic, the user must create a new topic, move the messages from the original topic to the new, and then delete the original. With this in mind, it is recommended to define logical spaces, based on an hierarchical topic naming structure. This setup can then be combined with security features, such as prefixed ACLs, to isolate different spaces and tenants, while also minimizing the administrative overhead for securing the data in the cluster.
+  </p>
+
+  <p>
+    These logical user spaces can be grouped in different ways, and the concrete choice depends on how your organization prefers to use your Kafka clusters. The most common groupings are as follows.
+  </p>
+
+  <p>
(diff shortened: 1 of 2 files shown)


miguno:
cc to committer @rajinisivaram as the SME on this subject, and who has the most context

bbejeck (approved):
LGTM and verified locally, but I'll wait on merging for a review from @rajinisivaram 

Also, @miguno, can you create an identical PR to go against docs in AK trunk?

miguno:
@bbejeck wrote in https://github.com/apache/kafka-site/pull/334#pullrequestreview-600990037:
> Also, @miguno, can you create an identical PR to go against docs in AK trunk?

Yes, I will do this once the content review of this PR is completed.

27/ops.html
```diff
@@ -1090,7 +1090,157 @@ <h4 class="anchor-heading"><a id="georeplication-monitoring" class="anchor-link"
...
+  <ul>
+    <li><code>&lt;project&gt;.&lt;product&gt;.&lt;event-name&gt;</code><br />(e.g., "mobility.payments.suspicious")</li>
+  </ul>
+
+  <p>
+    Certain information should normally not be put in a topic name, such as information that is likely to change over time (e.g., the name of the intended consumer) or that is a technical detail or metadata that is available elsewhere (e.g., the topic's partition count and other configuration settings).
+  </p>
+
+  <p>
+  To enforce a topic naming structure, it is useful to disable the Kafka feature to auto-create topics on demand by setting <code>auto.create.topics.enable=false</code> in the broker configuration. This stops users and applications from deliberately or inadvertently creating topics with arbitrary names, thus violating the naming structure. Then, you may want to put in place your own organizational process for controlled, yet automated creation of topics according to your naming convention, using scripting or your favorite automation toolkit.
```
dajac: I am not sure to get your point here. While I do agree that disabling the auto topic creation is a good thing, users/apps can still create topics with the admin client so it does not really help to enforce a topic naming structure. In both cases, the topics would have to respect the ACLs in place and the "namespace" if defined.

27/ops.html
```diff
@@ -1090,7 +1090,157 @@ <h4 class="anchor-heading"><a id="georeplication-monitoring" class="anchor-link"
...
...supports different types of (per-user principal) client quotas. Because a client's quotas apply irrespective of which topics the client is writing to or reading from, they are a convenient and effective tool to allocate resources in a multi-tenant cluster. <a href="#design_quotascpu">Request rate quotas</a>, for example, help to limit a user's impact on broker CPU usage by limiting the time a broker spends on the <a href="/protocol.html">request handling path</a> for that user, after which throttling kicks in. In many situations, isolating users with request rate quotas has a bigger impact in multi-tenant clusters than setting incoming/outgoing network bandwidth quotas, because excessive broker CPU usage for processing requests reduces the effective bandwidth the broker can serve. Furthermore, administrators can also define <a href="#brokerconfigs_controller.quota.window.num">quotas on topic operations</a> such as create, delete, and alter to prevent Kafka clusters from being overwhelmed by highly concurrent topic operations (see <a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-599%3A+Throttle+Create+Topic%2C+Create+Partition+and+Delete+Topic+Operations">KIP-599</a>).
```
dajac: `#brokerconfigs_controller.quota.window.num">` is not the correct config to highlight. `controller_mutations_rate` is more appropriate. We don't have an anchor for it though. I suggest to remove it for now.

I need to add more documentation about the controller quota. We can add a link to it here when it is done.

dajac (commented):
@miguno Thanks for the doc. This is great overall. I just left few small comments.

27/ops.html
```diff
@@ -1090,7 +1090,157 @@ <h4 class="anchor-heading"><a id="georeplication-monitoring" class="anchor-link"
...
+  </p>
+
+  <p>
+    <strong>Server quotas:</strong> In addition to client-side quotas, Kafka supports different types of broker-side quotas. For example, administrators can set a limit on the rate with which the <a href="#brokerconfigs_max.connection.creation.rate">broker accepts new connections</a>, set the <a href="#brokerconfigs_max.connections">maximum number of connections per broker</a>, or set the maximum number of connections allowed <a href="#brokerconfigs_max.connections.per.ip">from a specific IP address</a>.
```
rajinisivaram: Don't think we would refer to request rate or bandwidth quotas as `client-side`. They apply to clients, but are protecting broker resources and are enforced by brokers.

27/ops.html
rajinisivaram: Yes, the text sounds like you need to disable auto topic creation to enforce topic ACLs which is not the case, we should rewrite that part.

27/ops.html
```diff
@@ -1090,7 +1090,157 @@ <h4 class="anchor-heading"><a id="georeplication-monitoring" class="anchor-link"
...
+  </p>
+
+  <p>
+Security settings for Kafka fall into three main categories, which are similar to how administrators would secure other client-server data systems, like relational databases and traditional messaging systems.
+  </p>
+
+  <ol>
+    <li><strong>Encryption</strong> of data transferred between Kafka brokers and Kafka clients, between brokers, between brokers and ZooKeeper nodes, and between brokers and other, optional tools.</li>
+    <li><strong>Authentication</strong> of connections from Kafka clients and applications to Kafka brokers, as well as connections from Kafka brokers to ZooKeeper nodes.</li>
+    <li><strong>Authorization</strong> of client operations such as creating, deleting, and altering the configuration of topics; writing events to or reading events from a topic; creating and deleting ACLs.</li>
```
rajinisivaram: Should we also talk about `policies` like CreateTopicPolicy/AlterConfigPolicy that also support additional restrictions?

rajinisivaram (commented):
@miguno Thanks for the PR, looks good. Left a few minor comments.

27/ops.html
miguno: Ack and updated

27/ops.html
miguno: Yeah, I can add a note. But it's unfortunate that there's essentially zero coverage in the AK docs on how to use these.

27/ops.html
miguno: Ack and updated.

miguno:
PR updated with reviewer feedback.

dajac (approved):
LGTM, thanks!

rajinisivaram (approved):
@miguno Thanks for the updates, LGTM

miguno:
Thanks all for reviewing, much appreciated!

PR #345, opened 2021-04-05
Fix the formatting of example RocksDBConfigSetter

cc13ny:
https://issues.apache.org/jira/browse/KAFKA-12492

1. Fix the formatting of example RocksDBConfigSetter due to the un-arranged spaces within `<pre>` tag. 
2. Update it in the folder 27 so that the difference can be reviewed.
3. Duplicate the whole updated 27 folder as 28 folder so that it can be merged before the ongoing 2.8 is released.


ableegoldman:
Hey, thanks for the fix but please revert the 28 folder. For one thing, we don't want the 2.8 docs to go live until 2.8 is actually released. For another, this is just a copy of the 2.7 docs, the 2.8 docs are generated and copied over from the kafka repo during the release process. Also this may mess with the release script/release process, but I'm not sure about that. 

I just left a comment on the ticket but I'll reiterate it here: I would recommend to first submit a PR to fix this in the [kafka repo](https://github.com/apache/kafka) rather than in kafka-site, due to the ongoing 2.8 release. If the PR against the kafka repo can be merged in the next day or so, then it will get picked up in the 2.8 release and you won't even need to open a PR against the kafka-site repo at all. If we don't get it in before the 2.8 release, then you'll have to do the fix again in a PR against the kafka-site repo to fix it in the 28 folder that's generated during the 2.8 release

cc13ny:
Thanks. I created https://github.com/apache/kafka/pull/10486. I will close it now. But will reopen if it's not picked before the 2.8 release. Also how can I know if it will be picked?

ableegoldman:
>  Also how can I know if it will be picked?

I'll let you know :) 

But you can follow. along with the 2.8 release progress by subscribing to the dev mailing list. Look out for an email with a subject along the lines of `[VOTE] 2.8 RC1` , this means the RC for 2.8 has been cut and after that the release manager may proceed with the copy of docs from kafka/docs to kafka-site/28/docs. If your [PR](https://github.com/apache/kafka/pull/10486) hasn't been cherrypicked back to the 2.8 branch by then, then you'll need to reopen or submit a new PR against kafka-site. And of course I'll let you know on the PR once it's been merged, cherrypicked to 2.8, and so on.

PR #351, opened 2021-04-20
Add release info and download links for 2.6.2

ableegoldman:
To be merged once the release artifacts for 2.6.2 show up in Maven central

### DIFF
--- a/26/generated/connect_config.html
+++ b/26/generated/connect_config.html
@@ -375,7 +375,7 @@
 <table><tbody>
 <tr><th>Type:</th><td>list</td></tr>
 <tr><th>Default:</th><td>null</td></tr>
-<tr><th>Valid Values:</th><td>org.apache.kafka.connect.runtime.WorkerConfig$AdminListenersValidator@7b3300e5</td></tr>
+<tr><th>Valid Values:</th><td>org.apache.kafka.connect.runtime.WorkerConfig$AdminListenersValidator@d2cc05a</td></tr>
 <tr><th>Importance:</th><td>low</td></tr>
 </tbody></table>
 </li>
--- a/downloads.html
+++ b/downloads.html
@@ -102,6 +102,34 @@
         For more information, please read the detailed <a href="https://downloads.apache.org/kafka/2.7.0/RELEASE_NOTES.html">Release Notes</a>.
     </p>
 
+      <span id="2.6.2"></span>
+      <h3 class="download-version">2.6.2<a href="#2.6.2"><i class="fas fa-link " style="color:#053ce2"></i></a></h3>
+      <ul>
+          <li>
+              Released April 20, 2021
+          </li>
+          <li>
+              <a href="https://downloads.apache.org/kafka/2.6.2/RELEASE_NOTES.html">Release Notes</a>
+          </li>
+          <li>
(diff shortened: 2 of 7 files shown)


downloads.html
```diff
@@ -102,6 +102,34 @@ <h3 class="download-version">2.7.0<a href="#2.7.0"><i class="fas fa-link " style
...
+          <li>
+              <a href="https://downloads.apache.org/kafka/2.6.2/RELEASE_NOTES.html">Release Notes</a>
+          </li>
+          <li>
+              Source download: <a href="https://downloads.apache.org/kafka/2.6.2/kafka-2.6.2-src.tgz">kafka-2.6.2-src.tgz</a> (<a href="https://downloads.apache.org/kafka/2.6.2/kafka-2.6.2-src.tgz.asc">asc</a>, <a href="https://downloads.apache.org/kafka/2.6.2/kafka-2.6.2-src.tgz.sha512">sha512</a>)
+          </li>
+          <li>
+              Binary downloads:
+              <ul>
+                  <li>Scala 2.12 &nbsp;- <a href="https://downloads.apache.org/kafka/2.6.2/kafka_2.12-2.6.2.tgz">kafka_2.12-2.6.2.tgz</a> (<a href="https://downloads.apache.org/kafka/2.6.2/kafka_2.12-2.6.2.tgz.asc">asc</a>, <a href="https://downloads.apache.org/kafka/2.6.2/kafka_2.12-2.6.2.tgz.sha512">sha512</a>)</li>
```
ableegoldman: I found kind of a weird mix of different urls across the various releases on the downloads page (eg `https://www.apache.org/dyn/closer.cgi`), but presumably we want to point to `https://downloads.apache.org/kafka/2.6.2`?

rhauch: Actually, we want to follow that weird convention. I think the SHA512 and ASC files are small enough that we just use the downloads, but the other archive files are large enough that we should the `https://www.apache.org/dyn/closer.cgi` mechanism, which chooses a mirror that is closest to the user.

So let's change the links to be similar to the other recent releases.

rhauch:
Thanks, @ableegoldman. See my request above. Also, are there any documentation changes for the 2.6.2 release compared with 2.6.1? If so, we should include those in this PR.

26/generated/connect_config.html
```diff
@@ -375,7 +375,7 @@ <h4><a id="admin.listeners" href="#admin.listeners">admin.listeners</a></h4>
 <table><tbody>
 <tr><th>Type:</th><td>list</td></tr>
 <tr><th>Default:</th><td>null</td></tr>
-<tr><th>Valid Values:</th><td>org.apache.kafka.connect.runtime.WorkerConfig$AdminListenersValidator@7b3300e5</td></tr>
+<tr><th>Valid Values:</th><td>org.apache.kafka.connect.runtime.WorkerConfig$AdminListenersValidator@d2cc05a</td></tr>
```
ableegoldman, line 378: This is just some bug in the connect configs/docs, filed https://issues.apache.org/jira/browse/KAFKA-12700

rhauch: Thanks for dealing with this. Looks like a community member already submitted a PR, so we'll fix that shortly.

rhauch (approved):
LGTM. Thanks, @ableegoldman!

PR #352, opened 2021-04-21
Fix typo in Streams DSL API

edwardkenfox:
Found a simple typo in the Kafka Streams DSL in kafka-site. 

ref. https://kafka.apache.org/28/javadoc/org/apache/kafka/streams/kstream/SessionWindowedKStream.html

JFYI, I couldn't find the same typo in https://github.com/apache/kafka, so I believe this fix can/should be merged just by itself.


guozhangwang:
Thanks @edwardkenfox , I found in the kafka repo this is also typoed: https://github.com/apache/kafka/blob/trunk/docs/streams/developer-guide/dsl-api.html#L1077 will go ahead and fix it as well.

PR #373, opened 2021-09-15
Add Globo (https://www.globo.com/) to the list of the "Powered By ❤️"

allanassis:
I work at [Globo](https://globo.com/). We are a huge company here in Brazil and we use Kafka for our big data event ingestion and for our logging system.

Here how should appear at the powered-by page.

Captura de Tela 2021-09-15 às 19 33 23


bbejeck:
merged to asf-site thanks for the addition to the powered-by page

PR #374, opened 2021-09-20
Change 2.8.0 downloads to use archives

rhauch:
Now that 2.8.1 has been released, we can change the 2.8.0 download links to use the archive site.


mimaison (approved):
LGTM, thanks

PR #376, opened 2021-09-29
MINOR: remove note on future multi-cluster support of Kafka Streams

mjsax:
Follow up to https://github.com/apache/kafka/pull/11363


bbejeck (commented):
The lines changed here are different from the corresponding PR (https://github.com/apache/kafka/pull/11363), but otherwise this change LGTM.

mjsax:
Good catch! Wrong `Tip:` section.... Let me fix this.

PR #379, opened 2021-10-29
MINOR: Update coding guide to mention need to maintain public client API compatibility

stanislavkozlovski:
Previously, the coding guide would imply that it's acceptable to break API compatibility due to the project's early stage in maturity. For years now, the project has been mature and adopted enough to warrant not breaking API compatibility.


jolshan (approved):
This makes sense to me. I'm surprised we still had this wording on the site.

showuon (approved):
LGTM! That indeed doesn't apply to our current status.

mimaison (approved):
LGTM

PR #384, opened 2021-11-18
Update David Arthur photo

mumrah:
Current photo is about 10 years old, updating to a more recent one (also better quality).


jsancio (approved):
LGTM!

PR #387, opened 2021-12-08
Fixes KAFKA-13520

rmoff:
Quickstart currently fails on topic creation step because `--partitions` and `--replication-factor` are needed.

Ref: https://issues.apache.org/jira/browse/KAFKA-13520


bbejeck (approved):

bbejeck:
thanks for the fix @rmoff!

PR #388, opened 2021-12-14
adding new CVEs to list

scott-confluent:
Per https://docs.google.com/document/d/1zar4nlhIQnDB7qm9i4RqFUFXM5pWP7_a0WERcLlR0Wo/edit#heading=h.jp7xjzp1clqe

### DIFF
--- a/cve-list.html
+++ b/cve-list.html
@@ -21,7 +21,7 @@ This page lists all security vulnerabilities fixed in released versions of Apach
   <tbody>
     <tr>
       <td>Versions affected</td>
-      <td>NA</td>
+      <td>org.apache.logging.log4j:log4j-core &gt;=2.0-beta9 and &lt; 2.15.0</td>
     </tr>
     <tr>
       <td>Fixed versions</td>
@@ -41,7 +41,7 @@ This page lists all security vulnerabilities fixed in released versions of Apach
 <h2><a href="https://access.redhat.com/security/cve/CVE-2021-4104">CVE-2021-4104</a>
   Flaw in Apache Log4j logging library in versions 1.x</h2>
   
-  <p>Some components in Apache Kafka use <code>Log4j-v1.2.17</code></p>
+  <p>The following components in Apache Kafka use <code>Log4j-v1.2.17</code>: broker, controller, zookeeper, connect, mirrormaker and tools. Clients may also be configured to use <code>Log4j-v1.2.17</code>.</p>
 
   <p>Version 1.x of Log4J can be configured to use JMS Appender, which publishes log events to a JMS Topic. Log4j 1.x is vulnerable if the deployed application is configured to use JMSAppender.</p>
   
@@ -53,11 +53,18 @@ This page lists all security vulnerabilities fixed in released versions of Apach
     </tr>
     <tr>
       <td>Fixed versions</td>
-      <td>NA</td>
+      <td>
+        In the absence of a new log4j 1.x release, one can remove JMSAppender
+        from the log4j-1.2.17.jar artifact. Commands are listed in the
+        page <a href="http://slf4j.org/log4shell.html">http://slf4j.org/log4shell.html</a>.
+        <br />
+        <br />
+        We also recommend that configuration files be protected against write access as stated in <a href="http://slf4j.org/log4shell.html">http://slf4j.org/log4shell.html</a>.
+    </td>
     </tr>
     <tr>
       <td>Impact</td>
-      <td>This issue could result in remote code execution attack when the application is configured to use JMSAppender.</td>
+      <td>This issue could result in a remote code execution attack when the application is configured to use JMSAppender AND the attacker has access to directly modify the TopicBindingName or TopicConnectionFactoryBindingName configuration variables in property files which is typically an unlikely exploitation scenario.</td>
     </tr>
     <tr>
       <td>Issue announced</td>
--- a/cve-list.html
+++ b/cve-list.html
@@ -9,6 +9,63 @@
 
 This page lists all security vulnerabilities fixed in released versions of Apache Kafka.
 
+<h2><a href="https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-44228">CVE-2021-44228</a>
+  Flaw in Apache Log4j logging library in versions from 2.0.0 and before 2.15.0</h2>
+
+  <p>Some components in Apache Kafka use <code>Log4j-v1.2.17</code>  there is no dependence on <code>Log4j v2.*</code></p>
+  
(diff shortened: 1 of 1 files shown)


junrao (approved):
@scott-confluent : Thanks for the PR. LGTM

@showuon : Do you have any more comments on this?

scott-confluent:
We can't get a preview up, but here's the full page screenshot from my local:
screencapture-localhost-cve-list-html-2021-12-14-12_14_14

cve-list.html
```diff
@@ -9,6 +9,63 @@ <h1>Apache Kafka Security Vulnerabilities</h1>
...
+      <td>Issue announced</td>
+      <td>09 Dec 2021</td>
+    </tr>
+  </tbody>
+  </table>
+
+<h2><a href="https://access.redhat.com/security/cve/CVE-2021-4104">CVE-2021-4104</a>
+  Flaw in Apache Log4j logging library in versions 1.x</h2>
+  
+  <p>Some components in Apache Kafka use <code>Log4j-v1.2.17</code></p>
```
junrao: Could we spell out all components? I think the components include the following.

broker
controller
zookeeper
connect
mirrormaker
tools
clients configured with log4j

scott-confluent: How does something like this sound?

"Some components, including but not limited to: broker, controller, zookeeper, connect, mirrormaker, tools, and clients configured with log4j, use `Log4j-v1.2.17`"

cve-list.html
```diff
@@ -9,6 +9,63 @@ <h1>Apache Kafka Security Vulnerabilities</h1>
...
+  
+  <table class="data-table">
+  <tbody>
+    <tr>
+      <td>Versions affected</td>
+      <td>All versions</td>
+    </tr>
+    <tr>
+      <td>Fixed versions</td>
+      <td>NA</td>
```
junrao: Since we don't have a fix yet, perhaps we could add the following.

In the absence of a new log4j 1.x release, one can remove JMSAppender
from the *log4j-1.2.17.jar* artifact. Commands are listed in the
page <http://slf4j.org/log4shell.html>.

We also recommend that configuration files be protected against write access as stated in <http://slf4j.org/log4shell.html>.

cve-list.html
junrao: How about the following?

"The following components in Apache Kafka use Log4j-v1.2.17: broker, controller, zookeeper, connect, mirrormaker and tools. Clients may also be configured to use Log4j-v1.2.17."

scott-confluent: Here it is with both updates:
Screen Shot 2021-12-14 at 1 57 20 PM

cve-list.html
scott-confluent: Here it is with both updates:
Screen Shot 2021-12-14 at 1 57 20 PM

cve-list.html
junrao: A minor change. "Client applications may also be configured to use Log4j-v1.x."

junrao:
@scott-confluent : Have you pushed the new changes?

junrao:
@ijuma : Does the content look good to you? Thanks.

izzyacademy:
Do we want to add a disclaimer that users need to check their connectors to see if it uses log4j2?

Though the core library does not use this dependency, it is possible external connectors that use it could introduce vulnerabilities if they depend on the affected log4j2 version

scott-confluent:
> Do we want to add a disclaimer that users need to check their connectors to see if it uses log4j2?
> 
> Though the core library does not use this dependency, it is possible external connectors that use it could introduce vulnerabilities if they depend on the affected log4j2 version

@junrao what do you think?

scott-confluent:
@junrao I have pushed all updates from the doc as well as your comments here

cve-list.html
scott-confluent: Is this to update the last sentence of the first paragraph?

cve-list.html
```diff
@@ -9,6 +9,70 @@ <h1>Apache Kafka Security Vulnerabilities</h1>
...
+      <td>Issue announced</td>
+      <td>09 Dec 2021</td>
+    </tr>
+  </tbody>
+  </table>
+
+<h2><a href="https://access.redhat.com/security/cve/CVE-2021-4104">CVE-2021-4104</a>
+  Flaw in Apache Log4j logging library in versions 1.x</h2>
+  
+  <p>The following components in Apache Kafka use <code>Log4j-v1.2.17</code>: broker, controller, zookeeper, connect, mirrormaker and tools. Clients may also be configured to use <code>Log4j-v1.2.17</code>.</p>
```
junrao: Clients may also be configured to use <code>Log4j-v1.2.17 => Clients may also be configured to use <code>Log4j-v1.x

scott-confluent: Updated. Thanks for clarifying

scott-confluent:
screencapture-localhost-cve-list-html-2021-12-14-17_54_58

scott-confluent:
screencapture-localhost-cve-list-html-2021-12-14-18_22_24

rhauch:
As background for Connect: 
* Connect runtime puts all JARs from each connector plugin on a separate classloader, and the Connect runtime nor other connector plugins have access to a plugin's JARs. This is why a connector plugin that includes a Log4J 2.x JAR 
* Most connector implementations simply use the logging provided by the Connect runtime, which is Log4J 1.x regardless of the JARs included by connector plugins.
* However, if a connector plugins does include the Log4J 2.x JAR files, those JAR files will only be used if the connector implementation explicitly uses those APIs. There isn't a need to do this, but connectors are custom code and can do quite a bit.

We might consider adding something like this under the `CVE-2021-44228` section, which I hope conveys the limited scope of the risk:

> The Connect runtime of Apache Kafka allows users to install third party connector plugins. These connector plugins will use Connect runtime's Log4J 1.x by default, even when Log4J 1.x or 2.x JARs are inadvertently shipped with the connector plugin. Check with the vendor of any connector plugin that includes a Log4J 2.x JAR file.

Basically, AK is not responsible for third party connectors that users add to their Connect installations. But users should consult with the vendor of those third party connectors.

As for `CVE-2021-4104`, I think the existing wording applies to Connect just as well as every other part of AK, so IMO no changes are necessary to that section specifically for Connect.

Feel free to wordsmith as needed.

rhauch:
Or, if we want something even simpler, maybe we could just add a sentence to the `CVE-2021-44228` section that says:

>  Check with the vendor of any connector plugin that includes a Log4J 2.x JAR file.

junrao:
Thanks, @rhauch. The simpler version is probably enough. @scott-confluent : Could you add that?

scott-confluent:
@junrao @rhauch Does adding it here make sense?
Screen Shot 2021-12-14 at 6 59 27 PM

junrao:
@scott-confluent : Thanks for the updated PR. LGTM

PR #410, opened 2022-05-12
KAFKA-13882 Docker to preview docs locally

qingwei91:
I've added a Dockerfile and handy script to run documentation locally with 1 command.

It should reduce friction for docs contribution.


qingwei91:
I dont think we can do the same on https://github.com/apache/kafka/ itself, please correct me if I'm wrong

start-preview.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+docker build -t kafka-site-preview .
+
+docker run -dit --rm --name mypreview -p 8080:80 -v "$PWD":/usr/local/apache2/htdocs/ kafka-site-preview
+
+echo "You can stop the preview server by running `docker stop mypreview"
```
mimaison: Can we close the backtick and also add a new line?

README.md
```diff
@@ -0,0 +1,11 @@
...
+
+The documentation can hosted on a local webserver via httpd.
+
+You can run it with the following command, note that it requires docker:
+
+```shell
+sh start-preview.sh
+```
+
+Then you can open localhost:8080 on your browser and browse the documentation
```
mimaison: Can we make a link for localhost:8080?
Also let's add a full stop at the end of the sentence

README.md
```diff
@@ -0,0 +1,11 @@
+# How to preview the documentation changes locally?
+
+The documentation can hosted on a local webserver via httpd.
+
+You can run it with the following command, note that it requires docker:
+
+```shell
+sh start-preview.sh
```
mimaison: Can we make the script executable and call `./start-preview.sh`?

README.md
```diff
@@ -0,0 +1,11 @@
+# How to preview the documentation changes locally?
+
+The documentation can hosted on a local webserver via httpd.
```
mimaison: `The documentation can hosted` -> `The documentation can be hosted`

README.md
```diff
@@ -0,0 +1,11 @@
+# How to preview the documentation changes locally?
```
mimaison: We can drop `the` here

mimaison (commented):
Thanks for the PR. I left a few suggestions

README.md
qingwei91: Hi, does this means users have to run chmod +x?

mimaison: We don't want users to have to do it, make the file runnable (by running `chmod +x start-preview.sh`) as part of this PR

start-preview.sh
mimaison: Thanks for the update. Can we add a new line at the end of the file?

README.md
qingwei91: Got it, I've updated the file permission and the readme

start-preview.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+docker build -t kafka-site-preview .
+
+docker run -dit --rm --name mypreview -p 8080:80 -v "$PWD":/usr/local/apache2/htdocs/ kafka-site-preview
+
+echo "You can stop the preview server by running `docker stop mypreview`"
```
mimaison: What about doing something like this to make it even easier to use?

```suggestion
#!/usr/bin/env bash

set -euo pipefail

clean() {
  docker stop site-preview
  exit 0
}

trap clean SIGINT

docker build -t kafka-site-preview .

docker run -dit --rm --name site-preview -p 8080:80 -v "$PWD":/usr/local/apache2/htdocs/ kafka-site-preview

echo "The preview is available on http://localhost:8080."
echo "Press CTRL+C to stop it."

read -r -d '' _ </dev/tty

```

mimaison (commented):
Thanks for the updates. I left another suggestion to make the script even easier to use.

start-preview.sh
qingwei91: Hi, thanks for the suggestion, I am not a shell expert so I might be missing something here, I think we achieve similar effect by not running the container in detached mode.

Then ctrl + c will work, do you think that's acceptable?

qingwei91: ~Also, can you explain what `read -r -d '' _ </dev/tty` is trying to achieve?~

I figured it out, it is to block the process to not exit

mimaison: If you know a better way to achieve the same, yes do it

mimaison:
When I run this locally, I can only access pages under `Docs`, for anything else I get 404.

Is it expected? I would be nice to preview the complete website.

qingwei91:
Woops, might be a regression, sorry, let me check, I will move this into draft in the mean time

I agree, it is better to make it work for the whole site

qingwei91:
Hi @mimaison sorry for the back and forth, after some fiddling, I realize the problem is that when requesting `localhost:8080/path`, httpd tries to find a file named`path`, which does not exists, and what we really want is to resolve `path` to `path.html` in project root.

I solved this by adding a rewrite rule in `.htaccess` file, but I don't know how is it solved in the real website deployment, is `.htaccess` actually used in production?

.htaccess
```diff
@@ -9,3 +9,5 @@ RewriteRule ^/?(\d+)/javadoc - [S=2]
 RewriteRule ^/?(\d+)/images/ - [S=1]
 RewriteCond $2 !=protocol
 RewriteRule ^/?(\d+)/([a-z]+)(\.html)? /$1/documentation#$2 [R=302,L,NE]
+RewriteCond %{REQUEST_FILENAME}.html -f
+RewriteRule ^(.*)$ %{REQUEST_FILENAME}.html
```
qingwei91: Add a rewrite rule, it will serve `<REQUEST_FILENAME>.html` if the HTML file exists.

I do have a question, is this file being used in Production?

If its not, then this change should be fine, if it is, then maybe we need someone that is familiar with it to review, I never operate httpd so I might be missing something

qingwei91:
Hi @mimaison , if you get a chance, can you review this again? I've addressed the issue you pointed out, thanks!

.htaccess
mimaison: I find it strange we have to edit this file and I wonder how it works on the website. I'll try to find out

mjsax:
@mimaison -- seems this slipped. Can you help finishing it?

mimaison:
@mjsax The reason I did not merge is because I don't understand why when testing locally we need changes to the `.htaccess` file that don't seem necessary on the website. See the few comments starting from https://github.com/apache/kafka-site/pull/410#issuecomment-1126118797. Do you have an idea why this may be the case?

mjsax:
I just checkout the PR locally and it worked for me -- did you update your local copy after the latest changes?

qingwei91:
Hi @mjsax  and @mimaison 

After merging with master, I think the change in htaccess is no longer needed, I just tested it.
I've removed them

lucasbru:
In the meanwhile I redid the same _exact_ same thing here :P https://github.com/lucasbru/kafka-site-docker

mimaison:
@mjsax My point is that when testing locally without the `.htaccess` changes, all pages other than localhost:8080/documentation/ fail to load. For example http://localhost:8080/quickstart returns a 404.

I see that @qingwei91 removed the `.htaccess` changes in the last commit and I confirmed the issue happens again. I'm a bit scared of touching the `.htaccess` file as I don't want to break the real website.

qingwei91:
@mimaison , really sorry, I forgotten to check quick start. I've added it back now, guess I need to figure out why that's required here but not in the actual deployment ...

qingwei91:
Hi @mimaison , another way I can think of is to avoid using .htaccess in the docker, and inject all config required in httpd.conf instead.

https://github.com/apache/kafka-site/pull/410/commits/4d7397f17d10e64d3fadf4e8722e492bdf131fba

I am not sure if this is better since we now have duplicated config

I also wonder if .htaccess is actually used in the actual deployment, as it requires modifying httpd.conf

mjsax:
Ah sorry. Seems I miss-understood your comment @mimaison. -- I am not an expert in this domain and don't know why we need the change to `htaccess` nor what the actually impact is... :(

fvaleri (approved):
LGTM. Latest commit does not change .htaccess.

Thanks.

start-preview.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+docker build -t kafka-site-preview .
+
+docker run -it --rm --name mypreview -p 8080:80 -v "$PWD":/usr/local/apache2/htdocs/ kafka-site-preview
```
tombentley: I needed to add :z to the volume mount for this to work with podman (as opposed to docker) on Fedora)

```suggestion
docker run -it --rm --name mypreview -p 8080:80 -v "$PWD":/usr/local/apache2/htdocs/:z kafka-site-preview
```

tombentley:
@mimaison are you happy with the latest changes which don't touch the `.htaccess`? While I think the duplication is not _ideal_, this change would make it easier docs contributors.

divijvaidya:
Dropped by to request an update to https://cwiki.apache.org/confluence/display/KAFKA/Setup+Kafka+Website+on+Local+Apache+Server  and https://cwiki.apache.org/confluence/display/KAFKA/Contributing+Website+Documentation+Changes (the latter is linked from https://kafka.apache.org/contributing.html ) after this PR is merged.

mimaison:
@tombentley It would have been nice to understand the .htaccess inconsistency but the duplication is minor and this PR really simplifies testing the website so I'm happy to merge it.

@qingwei91 Can you integrate the change Tom suggested?

qingwei91:
Hi @mimaison , sure, I've incorporated the change. Would be nice to merge this if everything is ok

tombentley:
Thanks for the contribution @qingwei91! 

@divijvaidya thanks, updated.

PR #420, opened 2022-07-13
KAFKA-13868: Self host fonts with project website

divijvaidya:
As per the [Apache privacy policy](https://privacy.apache.org/faq/committers.html), Google Fonts are recommended to be hosted along with the website.

This change  adds the fonts locally in the code.

# Process
1. Determine the fonts that are used and the font weight using the existing css url `https://fonts.googleapis.com/css?family=Cutive+Mono|Roboto:100,300,400,700,900|Roboto+Condensed:300`
2. Determine the license of the font family from https://fonts.google.com/ 
3. Use https://google-webfonts-helper.herokuapp.com/fonts/ to download the fonts

# Testing
Tested the website locally to ensure sanity

**Before the change**
Screenshot 2022-07-13 at 13 21 20

**After the change (local)**
Screenshot 2022-07-13 at 13 19 31


divijvaidya:
@ijuma @mimaison please review.

css/fonts.css
```diff
@@ -0,0 +1,352 @@
+/* latin-ext */
+@font-face {
+    font-family: 'Cutive Mono';
+    font-style: normal;
+    font-weight: 400;
+    src: url(https://fonts.gstatic.com/s/cutivemono/v14/m8JWjfRfY7WVjVi2E-K9H6RMTm6o39ucNvc.woff2) format('woff2');
```
mimaison: Doesn't this mean browsers will still fetch fonts from Google's servers?

mimaison (commented):
As an alternative, have you tried getting rid of these custom fonts and using a standard one like helvetica?

css/fonts.css
divijvaidya: you are right. This is my bad! Will fix.

divijvaidya:
> As an alternative, have you tried getting rid of these custom fonts and using a standard one like helvetica?

No. Roboto is a fairly common and popular font used across websites. I don't want to change the look/feel of the website as part of this change.

I have added the font files now in this PR and also added the process of downloading them in the description. @mimaison, this is ready for your review.

showuon:
@scott-confluent , could you help review this PR since you added the google font into this project? Thanks.

showuon (commented):
@divijvaidya , why do you use https://google-webfonts-helper.herokuapp.com/fonts/ to download the fonts, instead of the original link: https://fonts.googleapis.com/css?family=Cutive+Mono|Roboto:100,300,400,700,900|Roboto+Condensed:300 ? Will that make any difference?

divijvaidya:
> @divijvaidya , why do you use https://google-webfonts-helper.herokuapp.com/fonts/ to download the fonts, instead of the original link: [https://fonts.googleapis.com/css?family=Cutive+Mono|Roboto:100,300,400,700,900|Roboto+Condensed:300](https://fonts.googleapis.com/css?family=Cutive+Mono%7CRoboto:100,300,400,700,900%7CRoboto+Condensed:300) ? Will that make any difference?

Hey @showuon 
If you visit the link https://fonts.googleapis.com/css?family=Cutive+Mono%7CRoboto:100,300,400,700,900%7CRoboto+Condensed:300, the `url` still points to Google's CDN and hence, does not fulfil our objective of self-hosting the fonts. An alternative could be to download the fonts from the source https://fonts.google.com/ but they only allow to download font family in `ttf` format. In that case, we would need to use a third party tool to convert this `ttl` to `woff2` compression. Further this site, https://fonts.google.com/, does not allow (at least I couldn't find it) to download for different charset such as vietnamese etc.

Note that helper tool I used https://google-webfonts-helper.herokuapp.com/fonts/  is using a MIT license and has [10K stars on Github](https://github.com/majodev/google-webfonts-helper), so I decided to trust it.

showuon:
@divijvaidya , I see. But in https://fonts.googleapis.com/css?family=Cutive+Mono%7CRoboto:100,300,400,700,900%7CRoboto+Condensed:300 , we can download all the fonts in `woff2` format via the link inside the font css, ex: 

```
/* latin-ext */
@font-face {
  font-family: 'Cutive Mono';
  font-style: normal;
  font-weight: 400;
  src: url(https://fonts.gstatic.com/s/cutivemono/v14/m8JWjfRfY7WVjVi2E-K9H6RMTm6o39ucNvc.woff2) format('woff2');
  unicode-range: U+0100-024F, U+0259, U+1E00-1EFF, U+2020, U+20A0-20AB, U+20AD-20CF, U+2113, U+2C60-2C7F, U+A720-A7FF;
}
```
We can download the font via https://fonts.gstatic.com/s/cutivemono/v14/m8JWjfRfY7WVjVi2E-K9H6RMTm6o39ucNvc.woff2 . 

However, I don't insist that we should download the font from google. I just want to make sure we don't miss any fonts we are using in Kafka website now. So, my next question is, how do we know we only need these 9 fonts you added? I saw there are 44 fonts css in this link: https://fonts.googleapis.com/css?family=Cutive+Mono%7CRoboto:100,300,400,700,900%7CRoboto+Condensed:300 . Does that mean we actually only use 9 of them? How do you know that?

Thank you again for helping working on this.

divijvaidya:
> We can download the font via https://fonts.gstatic.com/s/cutivemono/v14/m8JWjfRfY7WVjVi2E-K9H6RMTm6o39ucNvc.woff2

Oh yes. We can do that but I think that would be less maintainable. This is because when a new version of these fonts are available, we won't have a mechanism to update it. I can go ahead with this approach too if that makes it safer to accept this change. I don't have any strong opinions on this.

> Does that mean we actually only use 9 of them? How do you know that?

That is a very fair question to ask. The number of actual font files are lesser because the new font files contain all the charset in one file.

I will wait till tomorrow for others to add their thoughts. After that I can change this PR with the approach that you mentioned. Sounds ok Luke?

showuon:
> That is a very fair question to ask. The number of actual font files are lesser because the new font files contain all the charset in one file.

I see. Thanks.

> I will wait till tomorrow for others to add their thoughts. After that I can change this PR with the approach that you mentioned. Sounds ok Luke?

Yes, let's see what other people's thoughts. But I don't think it needs to adopt my approach if the font output is the same. 

Thank you.

css/fonts.css
```diff
@@ -0,0 +1,82 @@
+/* cutive-mono-regular - latin-ext_latin */
+@font-face {
+    font-family: 'Cutive Mono';
+    font-style: normal;
+    font-weight: 400;
+    src: local(''),
```
tombentley, line 6: Can you explain why the argument to `local` is the empty string? I'm no CSS expert, but from a quick google (https://stackoverflow.com/questions/3837249/font-face-src-local-how-to-use-the-local-font-if-the-user-already-has-it) it seems like it's supposed to be used for some kind of caching, but not with an empty string argument.

divijvaidya: https://stackoverflow.com/a/22835957 explains the rationale (for smiley but empty string follows same logic). Though this is only necessary for IE6-8 (very old browsers) to handle an edge case (different font with same name locally). I think I would remove it to reduce confusion while reading code.

scott-confluent (approved):
lgtm!

css/fonts.css
tombentley: Ah, thanks!

tombentley (approved):
I haven't actually tested it, but this LGTM.

showuon (approved):

mimaison (approved):

PR #423, opened 2022-07-13
Add a new signing key for David Arthur

mumrah:
This is in preparation for the upcoming 3.2.1 release. This key can also be found on [keys.openpgp.org](https://keys.openpgp.org/)


omkreddy (approved):
LGTM

PR #427, opened 2022-07-25
KAFKA-13868: Replace YouTube embedded video with hyperlinks

divijvaidya:
As per the [discussion in the community](https://lists.apache.org/thread/p24xvbf8nkvxpbj668vc0g3x3lojsnk4), we want to replace the embedded YouTube videos with hyperlinks to satisfy the [ASF privacy policy](https://privacy.apache.org/faq/committers.html).

This code change replaces the embedded videos from two of the pages on the website with hyperlinks.

**Before**
Screenshot 2022-07-25 at 19 31 50
Screenshot 2022-07-25 at 19 32 13

**After**
Screenshot 2022-07-25 at 19 32 01
Screenshot 2022-07-25 at 19 32 23

PR #430, opened 2022-07-29
MINOR: Add clickable images to load iframe videos

bbejeck:
This PR is a follow-up #427.  This PR adds a clickable image that will load an `iframe`;  in line with the [ASF privacy policy](https://privacy.apache.org/faq/committers.html) on embedded YouTube videos.  There will be a separate PR performing the same clickable image approach for the Kafka Streams videos.

### When the  introduction page loads:
intro_before_with_3rd_party_warning

### When the quickstart page loads:
qickstart_before_with_3rd_party_warning

### After clicking the image on the introduction page
after_introduction

### After clicking the image on the quickstart page
after-quickstart

I validated the behavior by running the Kafka site locally.

There are some additional times for follow-up:

1.  I am continuing to look into getting the videos hosted on the ASF YouTube page which will take care of the branding issue.
2. I've submitted a request to get the "What is Apache Kafka" video to remove the mention of Confluent.


bbejeck:
ping @divijvaidya, @mimaison , and @vvcephei

vvcephei (approved):
With my limited web knowledge, this looks like it'll get the job done to me!

One suggestion if it's not too much trouble, and if you agree... It might be obvious from context that that image is a button that plays a video, but it would be clearer if we superimposed the "play arrow" over the middle on the image. What do you think?

bbejeck:
>One suggestion if it's not too much trouble, and if you agree... It might be obvious from context that that image is a button that plays a video, but it would be clearer if we superimposed the "play arrow" over the middle on the image. What do you think?

I agree, but the "play arrow" on the loaded `iframe` is coming from YouTube.  I'm sure there's a way to do it, but that is beyond my CSS skills.  

I've updated the PR to have some text instructing to click to load the video.  The text disappears when a user clicks the image.

WDYT?

vvcephei:
Oh, that works! I was just thinking of editing the placeholder image itself to paste a stock play arrow on it.

bbejeck:
@vvcephei I took a stab at putting a play button on the image - I think this version wins.
Let me know what you think

mimaison:
Thanks @bbejeck. 

I tested it locally and it looks good. Can we do the same to the Streams videos on https://kafka.apache.org/32/documentation/streams/ ?

divijvaidya:
Thank you for fixing this @bbejeck!

Two questions:
1. How do we know that this code will work on all (or reasonable number of) browsers?

2. As per the [ASF guideline](https://privacy.apache.org/faq/committers.html), we also need to need to "Make it clear that users who click the image will load a video from a third party." How do we achieve this?

bbejeck:
> Thanks @bbejeck.
> 
> I tested it locally and it looks good. Can we do the same to the Streams videos on https://kafka.apache.org/32/documentation/streams/ ?

Hi @mimaison -yes I plan to have a PR submitted very soon (this week).  Actually, the Streams videos go back to the `10` version, so it will cover all of those as well

bbejeck:
> Thank you for fixing this @bbejeck!
> 
> Two questions:
> 
> 1. How do we know that this code will work on all (or reasonable number of) browsers?
> 2. As per the [ASF guideline](https://privacy.apache.org/faq/committers.html), we also need to need to "Make it clear that users who click the image will load a video from a third party." How do we achieve this?

Hi @divijvaidya - 

1. It's vanilla javascript, so any of today's standard browsers should support this
2. I can update the text to state as much.

vvcephei:
Thanks, @bbejeck ! My suggestion is addressed.

bbejeck:
@divijvaidya[ I found a free online JS compatibility tester ](https://seedmanc.github.io/jscc/)
Here is the result

Screen Shot 2022-08-01 at 1 57 18 PM

bbejeck:
Updated to include a warning about loading a video from a third party

divijvaidya:
> Updated to include a warning about loading a video from a third party

@bbejeck you might have forgotten to push the commit? 😅

divijvaidya (approved):
Thank you for making the changes @bbejeck. Looks good to me.

PR #441, opened 2022-09-14
Correct the year of release date for 2.7.1 in downloads.html

heciritoglu:
This is a super tiny thing, but could be misleading. I believe the release date year is not correct given 2.7.0 released later than 2.7.1 is a contradiction.  I also download the build to verify this, I can see the files are created in April 2021.


mjsax:
Thanks for the PR!

PR #446, opened 2022-09-28
MINOR: quick fix for TOC handle

mumrah:
Seen in the prerelease of the 3.3 docs, the TOC handle overlaps with the main contents of the `.right` div.

The left margin for this div was "auto", which was not getting calculated correctly. This patch sets the left margin to 33 rem which is what the actual left margin is supposed to be. Since the TOC and handle are fixed width, it's okay to use a fixed margin as well.

**Current 3.3 docs:**

----------------
image

----------------

**With this patch:**

----------------
image

----------------

There is likely some other issue in the body elements causing this margin to get miscalculated, though I couldn't find the root cause. 

Tested in Chrome 106.0.5249.61 and Safari 15.6.1 on Mac OS 12.5.1.

### DIFF
--- a/33/ops.html
+++ b/33/ops.html
@@ -1388,7 +1388,7 @@ $ bin/kafka-acls.sh \
 
   <pre class="line-numbers"><code class="language-bash"> &gt; bin/kafka-storage.sh format --cluster-id uuid --config server_properties</code></pre>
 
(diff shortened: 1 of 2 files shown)


bbejeck (approved):
The fix for the TOC handle is fixed

But there is a strange bar thing in the right - but This PR addresses the original problem so I'd say this LGTM and we can tackle the other thing separately

bbejeck:
Screen Shot 2022-09-28 at 5 04 04 PM

mumrah:
Somehow I completely missed that bar thing. I think that's the culprit here, let me see if I can track down where its coming from

bbejeck:
@mumrah the latest commit fixes it - thanks for the update

PR #480, opened 2023-01-19
MINOR: Add Stanislav Kozlovski to committer list

stanislavkozlovski:
This patch adds Stanislav Kozlovski to the committer page, including an image and a few social media links through which one can reach him (me)


committers.html
```diff
@@ -469,6 +469,18 @@ <h1 class="content-title">The committers</h1>
 
         </td>
       </tr>
+      <tr>
```
stanislavkozlovski, line 472: afaict we put a `<tr>` every two committers

stanislavkozlovski:
tested locally, looks like this: 
image

dajac (approved):
LGTM! Congrats!

PR #485, opened 2023-02-06
Add initial ZK to KRaft migration docs

mumrah:


ijuma (approved):
LGTM as an initial draft. Let's get this merged to go with the release and we can follow up with improvements after @cmccabe reviews.

PR #489, opened 2023-02-09
MINOR: explicitly set listener host name

showuon:
Workaround the error:
```
org.apache.kafka.common.errors.InvalidRequestException: Received request api key LEADER_AND_ISR which is not enabled
[2023-02-08 12:06:25,776] ERROR Exception while processing request from 192.168.1.11:9092-192.168.1.11:57210-107 (kafka.network.Processor)
```

by explicitly set listener host name


showuon:
@mimaison @mumrah , call for review

showuon:
I was wrong. It won't fix it. Please ignore it.

PR #493, opened 2023-02-19
Fixed typos in introduction documentation

anlance:


mimaison:
I believe we really mean `active/active` here. This means but clusters are actively used, compared to active/passive where only 1 of the clusters is used.

PR #499, opened 2023-03-30
Use ASF self service portal to get JIRA account

divijvaidya:
We no longer require users to ask the dev mailing list to setup a JIRA account.

### DIFF
--- a/contributing.html
+++ b/contributing.html
@@ -44,7 +44,7 @@
 
 		The easiest way to get started working with the code base is to pick up a really easy JIRA and work on that. This will help you get familiar with the code base, build system, review process, etc. We flag these kind of starter bugs <a href="https://issues.apache.org/jira/issues/?jql=project%20%3D%20KAFKA%20AND%20labels%20%3D%20newbie%20AND%20status%20%3D%20Open">here</a>.
 		<p>
-		Please contact us to be added to the contributor list with your JIRA account username provided in the email. After that you can assign yourself to the JIRA ticket you have started working on so others will notice.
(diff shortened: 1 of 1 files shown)


showuon (approved):
LGTM! Thanks for the update.

mimaison:
Don't people still need to ask for contributor permissions to self assign tickets?

divijvaidya:
> Don't people still need to ask for contributor permissions to self assign tickets?

Correct me if I am wrong but once you have a Apache JIRA account, you can self assign tickets. The ASF self serve workflows does not distinguish between user/contributor role.

mimaison:
In the past when people created accounts on their own, they could only create new tickets. We needed to assign contributor permissions so they could self assign tickets.

divijvaidya:
> In the past when people created accounts on their own, they could only create new tickets. We needed to assign contributor permissions so they could self assign tickets.

@mimaison You are right. See: https://issues.apache.org/jira/browse/INFRA-24423 

IMO, the committer/PMC approving the self-service request should follow-up with changing the permissions. Do we have a wiki/documentation for instructions to committers on how to do this? (similar to [this](https://tinkerpop.apache.org/docs/current/dev/developer/#_for_committers))

An alternate is to ask the user to send an email requesting for permissions but that just adds another step in the process and another set item of work for committers.

showuon:
> IMO, the committer/PMC approving the self-service request should follow-up with changing the permissions. 

That's what I usually did. After approving the request, add the user into the contributor list.

PR #269, opened 2020-06-08
Kafka nav and hompeage redesign

scott-confluent:

PR #504, opened 2023-04-21
KAFKA-10883: link to get started in the top navigation header

machi1990:
Closes https://issues.apache.org/jira/browse/KAFKA-10883

Hi @mimaison can you've a look at this?

### DIFF
--- a/includes/_top.htm
+++ b/includes/_top.htm
@@ -15,7 +15,7 @@
 			<nav class="top-nav" role="navigation" aria-label="Main menu">
 				<ul class="top-nav-list">
 					<li class="top-nav-item" role="menuitem">
-						<a href="#" class="top-nav-item-anchor" aria-haspopup="true" aria-expanded="false">
+						<a href="/get-started" class="top-nav-item-anchor" aria-haspopup="true" aria-expanded="false">
 							Get Started
 						</a>
 						<ul class="top-nav-menu" aria-hidden="true" role="menu" title="Get Started">


mimaison:
Hi @machi1990, thanks for the PR!

I agree that the behavior of the menu is a bit strange. The issue is that the "get started" page does not seem finished, so I wonder if it's the reason it's currently not reachable. 

For example, it says "All you need to know to get up and running with Apache Kafka can be found here. **Etc. Etc.**". Also the buttons don't seem to work.

So before enabling it, I think we should properly finish it. WDYT?

machi1990:
Hi @mimaison thanks for the review. You are right. I landed on this issue when I was trying to understand how the website is generated. I didn't appreciate that the other buttons don't work: when I navigated to the `/get-started` link and seeing the "Introduction" content, I wrongly assumed that all worked fine and we could quickly add the link and have the quick win. However, it seems that the "get-started" page hasn't been revisited since https://github.com/apache/kafka-site/pull/269 and it has always been the case that's it's not linked in the top navigation menu. I wonder if there are plans to rework the content, @scott-confluent  any thoughts ?

> So before enabling it, I think we should properly finish it. WDYT?

I'd agree, thanks. In this case, I think the JIRA https://issues.apache.org/jira/browse/KAFKA-10883 could be edited in the way it reflects this sentiment i.e the page needs to be properly finished and then linked in the top navigation menu. WDYT?

mimaison:
Yes feel free to edit the JIRA. I've granted you permissions so you should be able to self assign tickets if you want.

machi1990:
Thanks @mimaison I've edited the JIRA to capture what has been discussed here. I'll close this PR as it is also mentioned in the JIRA that once the page is redisigned/finished off it has to be linked.

PR #511, opened 2023-05-15
MINOR: Update collaborators to 10

vvcephei:
The wiki page documented a limit of 20 people in the collaborators list, but when we merged the change, we were notified that that limit is actually 10 people.

### DIFF
--- a/contributing.html
+++ b/contributing.html
@@ -89,7 +89,7 @@
 		</p>
 
 		<p>
-		In an effort to keep the Apache Kafka project running smoothly, and also to help contributors become committers, we have enabled these roles (See <a href="https://github.com/apache/kafka/blob/trunk/.asf.yaml">the Apache Kafka Infra config</a>). To keep this process lightweight and fair, we keep the list of contributors full by specifying the top N non-committers (sorted by number of commits they have authored in the last 12 months), where N is the maximum size of that list (currently 20). Authorship is determined by <code>git shortlog</code>. The list will be updated as part of the major/minor release process, three to four times a year.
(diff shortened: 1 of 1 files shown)


vvcephei:
See https://github.com/apache/kafka/pull/13722

vvcephei:
Hey @jlprat , do you mind reviewing this update as well? Thanks!

jlprat:
@vvcephei Now that you are on it, it seems the 2 confluence links are broken. Could you also fix those?

jlprat:
Strangely enough, I think I could navigate to those links before, but now they return a "page not found"

contributing.html
```diff
@@ -89,7 +89,7 @@ <h2>Collaborators</h2>
 		</p>
```
jlprat, line 89: The 2 links above this lines do not seem to work for me

jlprat (changes requested):
We should fix the links as well as they are broken at the moment

contributing.html
vvcephei: Hmm, the links still work for me, although you apparently have to be logged in to load that page (for unknown reasons).

jlprat: Ah! Right! I was logged out and I couldn't see it. If I'm logged in I can.

jlprat (approved):
LGTM! We need to think what to do with the links if they effectively are only for logged in users, we might want to add a note in parentheses on the links stating you need to be logged in to see them.

vvcephei:
That sounds good to me. I just added a small note about needing to be logged in. Thanks for the reviews!

jlprat (approved):
LGTM! Thanks @vvcephei

PR #519, opened 2023-06-06
MINOR: Add Kafka 3.4.1 documentation and javadoc and download link

showuon:
1. Generated do for 3.4.1
2. add download link for v3.4.1


34/javadoc/script.js
```diff
@@ -1,26 +1,26 @@
...
+ *
+ *
+ *
+ *
+ *
+ *
+ *
+ *
+ *
+ *
```
jlprat: I'm surprised about this one, but I'm guessing it's Oracle that changed the header on the file.

34/javadoc/search.js
```diff
@@ -1,26 +1,26 @@
...
+ *
+ *
+ *
+ *
+ *
+ *
+ *
+ *
+ *
+ *
```
jlprat: Same as above

jlprat (approved):
LGTM! Thanks @showuon

34/javadoc/script.js
showuon: Thanks @jlprat for pointing this out. I don't know why the generated output looks like this. I found in [2.0 doc](https://github.com/apache/kafka-site/blob/asf-site/20/javadoc/script.js), the result also looks like this. But I think we should be consistent with recent builds, i.e.  [3.5](https://github.com/apache/kafka-site/blob/asf-site/35/javadoc/script.js), which uses original license statement. I'll revert that change.

34/javadoc/search.js
showuon: Change reverted in https://github.com/apache/kafka-site/pull/519/commits/3b39b2a6b7c459ef9f53a3b5acf9ed1aedc39307

jlprat (approved):
Reapproving. Thanks @showuon

PR #554, opened 2023-10-04
MINOR: document how we deal with advisories for dependencies

raboof:
We get questions about advisories in dependencies on a regular basis, often in the form of security scanner tool output. Because of the high likelihood of false positives, the ASF policy is not to accept such reports as security issues without additional analysis: we consider the mere fact that an advisory exists for a dependency already public knowledge.
    
This update makes this clearer to reporters, helps them find out whether an advisory impacts Kafka for themselvels (by pointing to the dependency-check suppressions and the issue tracker), and calls on them to help out with analysis.

### DIFF
--- a/project-security.html
+++ b/project-security.html
@@ -35,6 +35,22 @@
 		<p>
                         For a list of security issues fixed in released versions of Apache Kafka, see <a href="/cve-list">CVE list</a>.
 		</p>
+		<h2>Advisories for dependencies</h2>
+		<p>
+			Many organizations use 'security scanning' tools to detect components for which advisories exist. While we generally encourage using such tools, since they are an important way users are notified of risks, our experience is that they produce a lot of false positives: when a dependency of Kafka contains a vulnerability, it is likely Kafka is using it in a way that is not affected. As such, we do not consider the fact that an advisory has been published for a Kafka dependency sensitive. Only when additional analysis suggests Kafka may be affected by the problem, we ask you to report this finding privately through <a href="mailto:security@kafka.apache.org?Subject=[SECURITY] My security issue" target="_top">security@kafka.apache.org</a>.
+		</p>
+		<p>
+			When handling such warnings, you can:
+			<ul>
+				<li>Check if our <a href="https://github.com/apache/kafka/blob/trunk/gradle/resources/dependencycheck-suppressions.xml">DependencyCheck suppressions</a> contain any information on this advisory.
+				<li>See if there is any discussion on this advisory in the <a href="https://issues.apache.org/jira/browse/KAFKA">issue tracker</a>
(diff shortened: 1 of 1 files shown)


bmscomp:
Thanks @raboof  , can you put a description for detailing the purpose of this pull request ?

project-security.html
```diff
@@ -35,6 +35,22 @@ <h1 class="content-title">Kafka security</h1>
 		<p>
                         For a list of security issues fixed in released versions of Apache Kafka, see <a href="/cve-list">CVE list</a>.
 		</p>
+		<h2>Advisories for dependencies</h2>
+		<p>
+			Many organizations use 'security scanning' tools to detect components for which advisories exist. While we generally encourage using such tools, since they are an important way users are notified of risks, our experience is that they produce a lot of false positives: when a dependency of Kafka contains a vulnerability, it is likely Kafka is using it in a way that is not affected. As such, we do not consider the fact that an advisory has been published for a Kafka dependency sensitive. Only when additional analysis confirms Kafka is affected by the problem, we ask you to report this finding privately through <a href="mailto:security@kafka.apache.org?Subject=[SECURITY] My security issue" target="_top">security@kafka.apache.org</a>.
```
divijvaidya: I would be more comfortable if we add something like "if you are unsure about impact to Kafka, err on the side of reporting". Asking this because users in the community may not be familiar with breadth of components in Kafka code base to make a call on impact.

divijvaidya:
cc: @mimaison @ijuma @showuon, I would also solicit your opinion on this one.

raboof:
> Thanks @raboof , can you put a description for detailing the purpose of this pull request ?

(updated the commit and PR message)

project-security.html
mimaison: +1

raboof: I'm happy to adapt the wording - of course it's up to you as PMC to decide how you want to deal with such cases.

For context: since anyone can easily run a dependency scanner, it's been our general policy to consider the mere fact that an advisory exists for a dependency already public knowledge. If a user is unsure about the impact, that doesn't really introduce any new information, so private reporting might not yet be necessary at that point.

I now changed the wording from "Only when additional analysis confirms Kafka is affected" to "Only when additional analysis suggests Kafka may be affected' - let me know if you'd like to see a stronger bias towards private reporting.

mjsax:
@raboof @divijvaidya -- What is the status of this PR? Seems nothing happened for a while. Do we still plant to get this merged, or should we close it?

divijvaidya:
I feel comfortable with the current wording on this PR. Will keep this open for another 3 days for others to chime in, else I will merge this one.

divijvaidya (approved):

PR #556, opened 2023-10-06
MINOR Update default docs point to 3.6.0 release docs

satishd:


downloads.html
```diff
@@ -6,12 +6,41 @@
...
 
     <p>
     You can verify your download by following these <a href="https://www.apache.org/info/verification.html">procedures</a> and using these <a href="https://downloads.apache.org/kafka/KEYS">KEYS</a>.
     </p>
 
+    <span id="3.6.0"></span>
+    <h3 class="download-version">3.6.0<a href="#3.6.0"><i class="fas fa-link " style="color:#053ce2"></i></a></h3>
+    <ul>
+        <li>
+            Released Oct 7, 2023
```
satishd: This is a temporary date, I will update it with the respective release date.

showuon (approved):

downloads.html
```diff
@@ -6,12 +6,41 @@
...
+        </li>
+        <li>
+            <a href="https://archive.apache.org/dist/kafka/3.6.0/RELEASE_NOTES.html">Release Notes</a>
```
divijvaidya: the latest version is usually downloaded from `downloads.apache.org` and all prior versions are moved to `archive.apache.org` as per https://www.apache.org/legal/release-policy.html#archived 

(you can view prior commits in this file for reference examples)

satishd: Good catch! Updated with the latest commit.

divijvaidya: Please move 3.5 artefacts to archive links as well.

satishd: I kept 3.5.1 still with `downloads.apache.org` as that is the latest release for 3.5 and it may have some updates later. We can change it to `archive.apache.org` once we decide we will not have any updates on 3.5. 

https://www.apache.org/legal/release-policy.html#when-to-archive
```
downloads.apache.org should contain the latest release in each branch that is currently under development. When development ceases on a version branch, releases of that branch should be removed from the project's download directory.
```

divijvaidya: makes sense

divijvaidya (approved):

downloads.html
satishd: Updated the release as Oct 10, 2023.

PR #545, opened 2023-09-18
Update ZK migration section to indicate GA

mumrah:
Also fix a few other small nitpicks I noticed

PR #560, opened 2023-10-13
MINOR Add note about KAFKA-15552 to 3.6 upgrade section

mumrah:


36/upgrade.html
```diff
@@ -84,7 +84,9 @@ <h5><a id="upgrade_360_kraft" href="#upgrade_360_kraft">Upgrading KRaft-based cl
 
     <h5><a id="upgrade_360_notable" href="#upgrade_360_notable">Notable changes in 3.6.0</a></h5>
     <ul>
-        <li>Apache Kafka now supports having both an IPv4 and an IPv6 listener on the same port. This change only applies to
+        <li>ZooKeeper to KRaft migrations are now recommended for production usage. One significant issue was found in the
+	  3.6.0 release which affects transactional producers https://issues.apache.org/jira/browse/KAFKA-15552.</li>
```
jolshan, line 88: we should clarify this is idempotent producers as well.

ijuma: And can we be clearer about what we're recommending? Are we saying that they should wait for 3.6.1 if they use idempotent producers (the default since 3.0) or transactions?

mimaison: This information is very important to communicate to our users. Today users that don't follow closely the mailing list or Jira (which is most users), all they know is that migration is now production ready and unfortunately they are likely to run into issues if they run it.
Can we rephrase slightly the sentence to address the comments from Justine and Ismael and push that to the website soon?

mjsax:
@mumrah -- what is the status of this PR? Was this also fixed in `kafka.git` (to avoid overwriting/reverting in case we would do another release -- even if unlikely for 3.6)

36/upgrade.html
mumrah: Consolidating this change with #545

PR #567, opened 2023-11-15
MINOR: Sort powered by page output alphabetically

dsmmcken:
I've tried to leave this page better than how I found it:

- For fairness, output is sorted alphabetically by root domain. The listing was quasi alphabetical, but looks like over time people have prepended and appended new companies into a now random order. Rather than re-sorting it once, and committing I've sorted the output by root domain (ie. eng.uber.com appears under U) such that list will always remain fair and orderly.
- A few logos had an excessively large file size. Anything over 100kb has been resized, removing 1.5mb of page weight with no loss in quality as re-sized image are still 2x the displayed size.
- Fixed a trailing space
- Added Deephaven to the list


dsmmcken:
Anyone available to review this?

jolshan:
The Cisco and Allegrograph images seem really tiny, but maybe they look ok on the page? Can you share a screenshot of the example? I can also pull this and run the pr code.

jolshan:
Screen Shot 2024-01-23 at 11 41 51 AM
Screen Shot 2024-01-23 at 11 41 40 AM

OK -- seems fine

dsmmcken:
Yeah, for some reason GitHub shows the sizes relative to the other one. So since the original was 7826px (!!), the new one at 320px looks tiny in the PR review.

powered-by.html
```diff
@@ -808,6 +813,22 @@ <h3>Want to appear on this page?</h3>
 
 <script>
     $(function () {
+        // enforce alphabetical sort by root domain for fairness
```
jolshan, line 816: I wonder if we are running this once per page load anyway if it is fair to randomize the logos. 

That could be considered more fair. Not sure if folks expect the same ordering on reload. I guess it could be hard to find a specific logo without searching directly.

dsmmcken: I looked at the list as being mostly alphabetical and accepted that as being the original intent. Would random order upset any search bots?

jolshan: That's fair. Perhaps this could be a separate discussion from this PR. 👍

jolshan (approved):

dsmmcken:
Thanks! FYI I don't have write access, so please merge if you are able.

PR #568, opened 2023-11-24
MINOR: Update docs for 3.6.1

mimaison:
I generated the javadoc with `OpenJDK Runtime Environment Temurin-17.0.2+8 (build 17.0.2+8)`.


36/javadoc/allclasses-index.html
```diff
@@ -11,7 +11,7 @@
 <link rel="stylesheet" type="text/css" href="script-dir/jquery-ui.min.css" title="Style">
 <link rel="stylesheet" type="text/css" href="jquery-ui.overrides.css" title="Style">
 <script type="text/javascript" src="script.js"></script>
-<script type="text/javascript" src="script-dir/jquery-3.6.1.min.js"></script>
```
jlprat, line 14: why the downgrade? How were the `3.6.0` javadoc files generated?

jlprat (commented):
A big chunk of the changes are a downgrade of jQuery, I wonder why this happened.
Also the couple of license files that change makes me think if we did the 3.6.0 update wrong or if this one is done differently than before.

36/javadoc/allclasses-index.html
mimaison: It's because the previous javadoc files were generated using a different JVM. I used `OpenJDK Runtime Environment Temurin-17.0.2+8 (build 17.0.2+8)`, while the previous version seems to have used (from the copyright/license files) an Oracle JVM.

jlprat: Makes sense

jlprat (approved):
LGTM, thanks @mimaison

PR #570, opened 2023-12-06
MINOR: Add 3.6.1 to downloads page

mimaison:


downloads.html
```diff
@@ -6,12 +6,40 @@
...
+                <li>Scala 2.13 &nbsp;- <a href="https://downloads.apache.org/kafka/3.6.1/kafka_2.13-3.6.1.tgz">kafka_2.13-3.6.1.tgz</a> (<a href="https://downloads.apache.org/kafka/3.6.1/kafka_2.13-3.6.1.tgz.asc">asc</a>, <a href="https://downloads.apache.org/kafka/3.6.1/kafka_2.13-3.6.1.tgz.sha512">sha512</a>)</li>
+            </ul>
+            We build for multiple versions of Scala. This only matters if you are using Scala and you want a version
+            built for the same Scala version you use. Otherwise any version should work (2.13 is recommended).
+        </li>
+    </ul>
+
+    <p>
+        Kafka 3.6.1 fixes 30 issues since the 3.6.0 release.
+        For more information, please read the detailed <a href="https://downloads.apache.org/kafka/3.6.1/RELEASE_NOTES.html">Release Notes</a>
```
jlprat: There won't be a blog post for this release? Checking previous releases, we had one also for patch versions (https://kafka.apache.org/blog#apache_kafka_351_release_announcement)

jlprat (approved):
LGTM the only slight note is that this version doesn't have a blog post (previous had one) but I don't think this is a blocker
Thanks @mimaison

downloads.html
mimaison: We used to only publish announcements to the Apache blog for major and minor releases. As we now have our own blog, it's easy to add an entry. I've pushed an update adding a short blog post.

jlprat (approved):
re approving just in case :)

Thanks @mimaison

PR #589, opened 2024-02-28
MINOR: Use archive repository for older releases

mimaison:
Before removing older releases from the svn repo, this updates their links to use the Apache archive instead of the mirrors.

### DIFF
--- a/downloads.html
+++ b/downloads.html
@@ -109,16 +109,16 @@
             Released Dec 11, 2023
         </li>
         <li>
-            <a href="https://archive.apache.org/dist/kafka/3.5.2/RELEASE_NOTES.html">Release Notes</a>
+            <a href="https://downloads.apache.org/kafka/kafka/3.5.2/RELEASE_NOTES.html">Release Notes</a>
         </li>
         <li>
-            Source download: <a href="https://archive.apache.org/dist/kafka/3.5.2/kafka-3.5.2-src.tgz">kafka-3.5.2-src.tgz</a> (<a href="https://archive.apache.org/dist/kafka/3.5.2/kafka-3.5.2-src.tgz.asc">asc</a>, <a href="https://archive.apache.org/dist/kafka/3.5.2/kafka-3.5.2-src.tgz.sha512">sha512</a>)
+            Source download: <a href="https://downloads.apache.org/kafka/kafka/3.5.2/kafka-3.5.2-src.tgz">kafka-3.5.2-src.tgz</a> (<a href="https://downloads.apache.org/kafka/kafka/3.5.2/kafka-3.5.2-src.tgz.asc">asc</a>, <a href="https://downloads.apache.org/kafka/kafka/3.5.2/kafka-3.5.2-src.tgz.sha512">sha512</a>)
         </li>
         <li>
             Binary downloads:
             <ul>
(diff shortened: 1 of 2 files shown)


downloads.html
```diff
@@ -109,16 +109,16 @@ <h3 class="download-version">3.5.2<a href="#3.5.2"><i class="fas fa-link " style
             Released Dec 11, 2023
         </li>
         <li>
-            <a href="https://downloads.apache.org/kafka/3.5.2/RELEASE_NOTES.html">Release Notes</a>
+            <a href="https://archive.apache.org/dist/kafka/3.5.2/RELEASE_NOTES.html">Release Notes</a>
```
divijvaidya: 3.5 branch is not EOL yet (it's first release will complete 12 months in June 2024). Why are we moving to archive for this?

divijvaidya (commented):
Please mention how we are interpreting Apache's instruction in the description here - https://infra.apache.org/release-distribution.html#archival 

I am assuming that our interpretation is:
1. If a minor version is within EOL, then it's latest bug fix version will *not* be in archive since it is considered an active branch
2. All EOL minor versions are on archive
3. For non-EOL minor version, all bug fixe releases prior to latest will use *archive*

downloads.html
mimaison: Apache asks us to remove older releases from the mirroring system to reduce the load. Currently we have the following releases in the mirroring system: 3.4.1, 3.5.0, 3.5.1, 3.5.2, 3.6.0, 3.6.1 and 3.7.0.

So I'm doing a bit of cleanup as it seems it was missed in previous releases. As per the link you shared above:

> Each project's distribution directory should contain the latest release in each branch that is currently under development. When development ceases on a version branch, the PMC should remove links to releases of that branch from their download directory.

While we still support 3.5.2, I wouldn't say it's currently under development and no new releases for that branch are planned. So I was planning on deleting 3.4.1, 3.5.0, 3.5.1, 3.5.2 and 3.6.0 from the mirroring system. The releases are still downloadable via the archive repos and via maven. Also as 3.5.2 is below 3.6.0, which should definitively be removed from the mirroring system, in the downloads page, I thought it was just simpler to switch to the archive links for 3.5.2.

If you have a strong objection I'm happy to keep 3.5.2 in the mirroring system for now. We have a tendency to forget to clean it up, as you can see in https://issues.apache.org/jira/browse/KAFKA-6223, so each time we delete quite a few versions.

divijvaidya: I don't have an objection but I want to ensure that we have some fixed criteria on what to move to archive and what not to. I agree it's unlikely that we will have a bug fix version of 3.5.x until it is EOL in June but technically, it's still not EOL'ed.

I am ok with the criteria being, best judgement of the PMC member at the time of release of a minor version. My motivation is to ensure that we have a criteria which all PMC members could follow in future.

> We have a tendency to forget to clean it up

Is there something we can do in the release process to enforce this? I was thinking of having a multi-stage script which is run by PMC member and it will prompt whether this is done or not (or automatically do it). We can take the help of community to improve this as part of https://issues.apache.org/jira/browse/KAFKA-15198

mimaison: My understanding is that we use "best judgment" to select what to delete so it may vary each time. Maybe deciding to keep all not EOL'ed releases in the mirroring system would be a better process. 
I'll undo the 3.5.2 changes and will keep this release in the mirroring system for now.

> Is there something we can do in the release process to enforce this?
Some steps have to be performed by PMC members and once a release is announced it's easy to move onto something else and not complete the post release steps. If possible, yes we should try to automate as much as possible but it's often tricky.

mimaison: I've undone the changes related to 3.5.2.

divijvaidya:
We need a similar change in blog as well please (e.g. https://github.com/apache/kafka-site/blob/7ae222e3a5a80252063f7bd068cada382b39812b/blog.html#L317) otherwise we will have inconsistent broken links in the blog section.

mimaison:
Good catch! I pushed an update.

divijvaidya (approved):
Thank you making this change!

PR #597, opened 2024-04-11
[MINOR] Add youtube to content security policy

VedarthConfluent:
This change should fix the broken video on quickstart page https://kafka.apache.org/quickstart.
Currently observing following error:-
`quickstart:297 Refused to frame 'https://www.youtube.com/' because it violates the following Content Security Policy directive: "frame-src 'self'"`


omkreddy (approved):
LGTM

raboof:
unfortunately it seems https://w3c.github.io/webappsec-csp/#meta-element is correct that the meta element can only further restrict the policy, not widen it.

I wonder if you could do a `Header add` in a `.htaccess` file like in https://github.com/apache/privacy-website/tree/main/examples/youtube-html ?

PR #600, opened 2024-05-02
MINOR: Add code signing key for Igor

soarez:
For the first PMC reviewing this, could you please update the KEYS file in https://dist.apache.org/repos/dist/release/kafka/KEYS and leave a comment in this PR?


showuon (approved):
LGTM!

PR #603, opened 2024-06-11
MINOR: Add Josep Prat Key

jlprat:
Once merged, could a PMC member update the KEYS file in https://dist.apache.org/repos/dist/release/kafka/KEYS?

Thanks!


soarez (approved):
It looks like your key doesn't have the `CODE SIGNING KEY` comment. https://infra.apache.org/release-signing.html#key-comment

Otherwise LGTM

jlprat:
I see that others also lack this comment? How important is it to have it?

soarez:
I noticed the same. I don't know if it is actually important. I suppose you can add another key later if it proves really necessary.

mimaison:
I added the key to https://dist.apache.org/repos/dist/release/kafka/KEYS

jlprat:
Thanks @mimaison !

PR #606, opened 2024-06-12
MINOR: Mention confirming email subscriptions to contact page

gharris1727:
Some users partially subscribe to the mailing list, possibly because we don't mention the second step of replying to the confirmation email. Mentioning this succinctly in the page and linking to the ASF documentation could help make this less likely.


ableegoldman (approved):
LGTM, thanks for the improvement

PR #612, opened 2024-06-28
Add Christo to committers

clolov:
Just adding my face to the list 😊


committers.html
```diff
@@ -597,6 +597,18 @@ <h1 class="content-title">The committers</h1>
           <a href="https://github.com/soarez">github.com/soarez</a><br>
         </td>
       </tr>
+      <tr>
```
mimaison: Instead of creating a new row, you can put your picture in a second cell on the right of Igor.
Otherwise it looks like
image

You can use `./start-preview.sh` to visualize changes.

mimaison (approved):

PR #614, opened 2024-07-16
Add blog post for 3.8 release

jlprat:
image

### DIFF
--- a/38/upgrade.html
+++ b/38/upgrade.html
@@ -19,7 +19,7 @@
 
 <script id="upgrade-template" type="text/x-handlebars-template">
 
-<h4><a id="upgrade_3_8_0" href="#upgrade_3_8_0">Upgrading to 3.8.0 from any version 0.8.x through 3.6.x</a></h4>
+<h4><a id="upgrade_3_8_0" href="#upgrade_3_8_0">Upgrading to 3.8.0 from any version 0.8.x through 3.7.x</a></h4>
 
     <h5><a id="upgrade_380_notable" href="#upgrade_380_notable">Notable changes in 3.8.0</a></h5>
     <ul>
--- a/38/upgrade.html
+++ b/38/upgrade.html
@@ -28,7 +28,8 @@
             If READ is not authorized, checkpointing is limited to offsets mirrorred after the start of the task.
             See <a href="https://issues.apache.org/jira/browse/KAFKA-15905">KAFKA-15905</a> for more details.
         </li>
-        <li>Tiered Storage now supports clusters configured with multiple log directories (i.e. JBOD feature).</li>
+        <li>JBOD in KRaft is now production ready.</li>
+        <li>Tiered Storage, which is still in early availability, now supports clusters configured with multiple log directories (i.e. JBOD feature).</li>
     </ul>
 
 <h4><a id="upgrade_3_7_1" href="#upgrade_3_7_1">Upgrading to 3.7.1 from any version 0.8.x through 3.6.x</a></h4>
--- a/38/upgrade.html
+++ b/38/upgrade.html
@@ -29,7 +29,7 @@
             See <a href="https://issues.apache.org/jira/browse/KAFKA-15905">KAFKA-15905</a> for more details.
         </li>
         <li>JBOD in KRaft is now production ready.</li>
-        <li>Tiered Storage, which is still in early availability, now supports clusters configured with multiple log directories (i.e. JBOD feature).</li>
+        <li>Tiered Storage, which is still in early access, now supports clusters configured with multiple log directories (i.e. JBOD feature).</li>
     </ul>
 
 <h4><a id="upgrade_3_7_1" href="#upgrade_3_7_1">Upgrading to 3.7.1 from any version 0.8.x through 3.6.x</a></h4>
--- a/38/upgrade.html
+++ b/38/upgrade.html
@@ -28,7 +28,7 @@
             If READ is not authorized, checkpointing is limited to offsets mirrorred after the start of the task.
             See <a href="https://issues.apache.org/jira/browse/KAFKA-15905">KAFKA-15905</a> for more details.
         </li>
-        <li>JBOD in KRaft is now production ready.</li>
+        <li>JBOD in KRaft is no longer in early access.</li>
         <li>Tiered Storage, which is still in early access, now supports clusters configured with multiple log directories (i.e. JBOD feature).</li>
     </ul>
 
--- a/blog.html
+++ b/blog.html
@@ -43,6 +43,10 @@
                     Up until now, only the default compression level was used by Apache Kafka. From this version on, a configuration mechanism to specify compression level is included. See <a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-390%3A+Support+Compression+Level">KIP-390</a> for more details.
                 </p>
                 <p>
+                    In the last release, 3.7, <a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol">KIP-848 The Next Generation of the Consumer Rebalance Protocol</a> was made available as early access. This version includes numerous bug
+                    fixes and the community is encouraged to test and provide feedback. <a href="https://cwiki.apache.org/confluence/display/KAFKA/The+Next+Generation+of+the+Consumer+Rebalance+Protocol+%28KIP-848%29+-+Early+Access+Release+Notes">See the early access release notes for more information.</a>
+                </p>
+                <p>
                     The configuration value <code>offsets.commit.required.acks</code> is deprecated in this version and it will be removed in Kafka 4.0. See <a href="https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=303794933">KIP-1041</a> for more details.
                 </p>
                 <p>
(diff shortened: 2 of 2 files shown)


soarez (approved):
LGTM

jlprat:
The contributors list (and number) seems high, but I double checked the list and removed duplicates.
This is the command I run:
```
git shortlog -sn --group=author --group=trailer:co-authored-by --group=trailer:Reviewers --no-merges 3.7.0..3.8.0-rc1 | cut -f2 | sort --ignore-case | uniq | sed -e ':a' -e '$!N;s/\n/, /;ta' -e 's/,$//' -e 's/%$//'
```

blog.html
```diff
@@ -22,6 +22,94 @@
...
+                <p>See the <a href="https://kafka.apache.org/documentation.html#upgrade_3_8_0">Upgrading to 3.8.0 from any version 0.8.x through 3.6.x</a> section in the documentation for the list of notable changes and detailed upgrade steps.</p>
+                <p>
+                    In the last release, 3.6,
+                    <a href="https://kafka.apache.org/38/documentation.html#tiered_storage">tiered storage</a> was released as early availability feature.
+                    In this release, Tiered Storage now supports clusters configured with multiple log directories (i.e. JBOD feature).
+                </p>
+                <p>
+                    Up until now, only the default compression level was used by Apache Kafka. From this version on, a configuration mechanism to specify compression level is included. See <a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-390%3A+Support+Compression+Level">KIP-390</a> for more details.
```
stanislavkozlovski, line 43: Shall we add some perf numbers as shown by the tests ran in the KIP? 56.5% better message/s throughput & 40% latency reduction at the cost of 9% more disk usage in the gzip example is definitely worthy of a call out. Even more so the lz4 example. 

Unless I'm missing something, this is a very impactful config for users to change, and it's very easy. We should urge them to explore it

blog.html
```diff
@@ -22,6 +22,94 @@
...
+            <article>
+                <h2 class="bullet">
+                    <a id="apache_kafka_380_release_announcement"></a>
+                    <a href="#apache_kafka_380_release_announcement">Apache Kafka 3.8.0 Release Announcement</a>
+                </h2>
+                xx July 2024 - Josep Prat (<a href="https://twitter.com/jlprat">@jlprat</a>)
+                <p>We are proud to announce the release of Apache Kafka 3.8.0. This release contains many new features and improvements. This blog post will highlight some of the more prominent features. For a full list of changes, be sure to check the <a href="https://downloads.apache.org/kafka/3.8.0/RELEASE_NOTES.html">release notes</a>.</p>
+                <p>See the <a href="https://kafka.apache.org/documentation.html#upgrade_3_8_0">Upgrading to 3.8.0 from any version 0.8.x through 3.6.x</a> section in the documentation for the list of notable changes and detailed upgrade steps.</p>
+                <p>
+                    In the last release, 3.6,
```
stanislavkozlovski: 3.7 is the last release

blog.html
```diff
@@ -22,6 +22,94 @@
...
         <div class="right">
             <h1 class="content-title">Blog</h1>
+            <article>
+                <h2 class="bullet">
+                    <a id="apache_kafka_380_release_announcement"></a>
+                    <a href="#apache_kafka_380_release_announcement">Apache Kafka 3.8.0 Release Announcement</a>
+                </h2>
+                xx July 2024 - Josep Prat (<a href="https://twitter.com/jlprat">@jlprat</a>)
+                <p>We are proud to announce the release of Apache Kafka 3.8.0. This release contains many new features and improvements. This blog post will highlight some of the more prominent features. For a full list of changes, be sure to check the <a href="https://downloads.apache.org/kafka/3.8.0/RELEASE_NOTES.html">release notes</a>.</p>
+                <p>See the <a href="https://kafka.apache.org/documentation.html#upgrade_3_8_0">Upgrading to 3.8.0 from any version 0.8.x through 3.6.x</a> section in the documentation for the list of notable changes and detailed upgrade steps.</p>
```
stanislavkozlovski: through 3.7?

blog.html
```diff
@@ -22,6 +22,94 @@
...
+                <p>
+                    In the last release, 3.6,
+                    <a href="https://kafka.apache.org/38/documentation.html#tiered_storage">tiered storage</a> was released as early availability feature.
```
stanislavkozlovski, line 35: TS **with JBOD** was released as an EA feature. Are we now saying TS wit JBOD is GA?

blog.html
jlprat: True, but in this case, we need to say, in a previous release. I'll fix this

blog.html
jlprat, line 35: Makes sense

blog.html
```diff
@@ -22,6 +22,94 @@
...
+                    In this release, Tiered Storage now supports clusters configured with multiple log directories (i.e. JBOD feature).
```
stanislavkozlovski: Which KIP talks about this?

I know KRaft + JBOD was released as EA last release - 3.7. Is it GA today?

How does tiered storage relate?

blog.html
jlprat: Good catch. The documentation is wrong in `trunk` and `3.8`. I'll fix those.

blog.html
jlprat: I'm repeating what's in the release notes for `3.8`. Maybe this is wrong. Could someone clarify? https://github.com/apache/kafka/blob/trunk/docs/upgrade.html#L31

blog.html
jlprat: Where can I find the results? So I can properly link them in the blog

blog.html
soarez: Good point @stanislavkozlovski.

> I know KRaft + JBOD was released as EA last release - 3.7. Is it GA today?

I think it's fair to say JBOD in KRaft is GA in 3.8. 

Tiered Storage support for JBOD didn't get a KIP, it was implemented in KAFKA-16399.

blog.html
stanislavkozlovski: https://cwiki.apache.org/confluence/display/KAFKA/KIP-390%3A+Support+Compression+Level
image

blog.html
stanislavkozlovski: > I think it's fair to say JBOD in KRaft is GA in 3.8.

Agreed. It seems like I got confused with calling JBOD GA due to Tiered Storage. For context:

When I released 3.7, I was only aware of these two limitations described here - https://cwiki.apache.org/confluence/display/KAFKA/Kafka+JBOD+in+KRaft+Early+Access+Release+Notes
Both of those are NOT Tiered Storage related

AFAICT, Tiered Storage itself **is still in Early Access**. It was introduced in 3.6 in EA, 3.7 didn't change anything.
So if both JBOD in KRaft and TS are early access, then they probably don't block each other.

What I understand here, is the following:
- JBOD in KRaft **is GA** in 3.8, and this is due to some other changes
- Tiered Storage now supports JBOD as well (still EA)

stanislavkozlovski: We should probably mention both separately and explicitly.

Further, we should probably think of a way to track in-flight GAs. We have a few features that were in EA in parallel - tiered storage, the new consumer group protocol, JBOD in KRaft. It's hard to track the status of these things.

Not to mention the need to define what EA even means in the context of Apache Kafka :)

blog.html
mimaison: The improvement (or not) really depends on your data. The dataset mentioned in the KIP was in JSON which really benefits from compression.

jlprat: As the entry already mentions "See KIP-390 for more details" I think we are then covered. Unless we want to make the blog post dive deeper into showcasing a particular example of a benchmark of this feature.

blog.html
jlprat: @stanislavkozlovski Are we sure JBOD in KRaft is not GA?

jlprat: I'm asking because it has been a bit of a mystery how features are determined production ready in Apache Kafka.

soarez: My understanding, especially wrt JBOD in KRaft, is that we may flag new features as EA if we release them with serious bugs or limitations that users may find surprising.

I think the concept of "production ready", or GA, arises naturally from the negation of the term EA. I don't know if these terms should mean anything more than just "not EA".

jlprat: pushed a fix

38/upgrade.html
```diff
@@ -28,7 +28,8 @@ <h5><a id="upgrade_380_notable" href="#upgrade_380_notable">Notable changes in 3
             If READ is not authorized, checkpointing is limited to offsets mirrorred after the start of the task.
             See <a href="https://issues.apache.org/jira/browse/KAFKA-15905">KAFKA-15905</a> for more details.
         </li>
-        <li>Tiered Storage now supports clusters configured with multiple log directories (i.e. JBOD feature).</li>
+        <li>JBOD in KRaft is now production ready.</li>
+        <li>Tiered Storage, which is still in early availability, now supports clusters configured with multiple log directories (i.e. JBOD feature).</li>
```
jlprat: @stanislavkozlovski and @soarez I updated the `upgrade.html` file as well in here so the docs are fixed. If you like this I'll update https://github.com/apache/kafka/pull/16654 accordingly

blog.html
```diff
@@ -22,6 +22,98 @@
...
+                <p>See the <a href="https://kafka.apache.org/documentation.html#upgrade_3_8_0">Upgrading to 3.8.0 from any version 0.8.x through 3.7.x</a> section in the documentation for the list of notable changes and detailed upgrade steps.</p>
+                <p>
+                    In a previous release, 3.6,
+                    <a href="https://kafka.apache.org/38/documentation.html#tiered_storage">tiered storage</a> was released as early availability feature.
+                    In this release, Tiered Storage now supports clusters configured with multiple log directories (i.e. JBOD feature). This feature still remains as early availability.
+                </p>
+                <p>
+                    In the last release, 3.7, <a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-858%3A+Handle+JBOD+broker+disk+failure+in+KRaft">KIP-858</a>
+                    was released in early availability. Since this version, JBOD in KRaft is considered production ready.
```
soarez: The term used in the 3.7 release was _early access_. Would it be clearer to keep the same term?

What do you think of this variation?

```suggestion
                    <a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-858%3A+Handle+JBOD+broker+disk+failure+in+KRaft">KIP-858</a> introduced JBOD support in KRaft in the 3.7 release. Since then, limitations have been addressed and it is no longer an early access feature.
```

blog.html
stanislavkozlovski: @soarez - that's my understanding too

38/upgrade.html
soarez: Slight preference for using the "early access" term instead of "production ready", like in my suggestion in https://github.com/apache/kafka-site/pull/614#discussion_r1687161313 as I suspect there's less misunderstanding that way.

But I think this is OK. Thanks @jlprat

jlprat: Hi @soarez I think you are looking at an outdated version of the text. The new one states:
> <li>Tiered Storage, which is still in early access, now supports clusters configured with multiple log directories (i.e. JBOD feature).</li>

soarez: Ah, maybe I was looking at an outdated version. 

Can you also replace the terms in the `38/upgrade.html` file?
https://github.com/apache/kafka-site/pull/614/files#diff-0846773f713928726a99b67a3aa47f827fe120106c3e002fd28ca767f9e9efd3R31

https://github.com/apache/kafka-site/blob/547328dfdfcad9a25e2e2b3bee182924ec786c41/38/upgrade.html#L31

jlprat: done

blog.html
```diff
@@ -22,6 +22,98 @@
         <!--#include virtual="includes/_nav.htm" -->
         <div class="right">
             <h1 class="content-title">Blog</h1>
+            <article>
+                <h2 class="bullet">
+                    <a id="apache_kafka_380_release_announcement"></a>
+                    <a href="#apache_kafka_380_release_announcement">Apache Kafka 3.8.0 Release Announcement</a>
+                </h2>
+                xx July 2024 - Josep Prat (<a href="https://twitter.com/jlprat">@jlprat</a>)
```
soarez: Don't forget to update the day here.

soarez (approved):
LGTM thanks @jlprat

blog.html
```diff
@@ -22,6 +22,106 @@
...
+                    was released in early access. Since this version, JBOD in KRaft is no longer considered an early access feature.
+                </p>
+                <p>
+                    Up until now, only the default compression level was used by Apache Kafka. From this version on, a configuration mechanism to specify compression level is included. See <a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-390%3A+Support+Compression+Level">KIP-390</a> for more details.
+                </p>
+                <p>
+                    In the last release, 3.7, <a href="https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol">KIP-848 The Next Generation of the Consumer Rebalance Protocol</a> was made available as early access. This version includes numerous bug
+                    fixes and the community is encouraged to test and provide feedback. <a href="https://cwiki.apache.org/confluence/display/KAFKA/The+Next+Generation+of+the+Consumer+Rebalance+Protocol+%28KIP-848%29+-+Early+Access+Release+Notes">See the early access release notes for more information.</a>
```
jlprat, line 47: Hi @kirktrue, please take a look and let me know if this is what you had in mind. Please note that Apache Kafka hasn't defined yet properly what "early access" means. And we have no concept of "preview" or anything similar yet.

kirktrue: Yes. This seems fine. Thank you!

PR #620, opened 2024-08-01
KAFKA-17146 Include note to remove migration znode

mumrah:
When reverting the ZK migration, we must also remove the `/migration` ZNode in order to allow the migration to be re-attempted in the future.

### DIFF
--- a/38/ops.html
+++ b/38/ops.html
@@ -4145,7 +4145,10 @@ listeners=CONTROLLER://:9093
             </li>
             <li>
               Using <code>zookeeper-shell.sh</code>, run <code>rmr /controller</code> so that one
-              of the brokers can become the new old-style controller.
+              of the brokers can become the new old-style controller. Additionally, run
+              <code>get /migration</code> followed by <code>rmr /migration</code> to clear the
+              migration state from ZooKeeper. This will allow you to re-attempt the migration
+              in the future. The data read from "/migration" can be useful for debugging.
             </li>
             <li>
               On each broker, remove the <code>zookeeper.metadata.migration.enable</code>,
--- a/38/ops.html
+++ b/38/ops.html
@@ -4105,7 +4105,10 @@ listeners=CONTROLLER://:9093
             </li>
             <li>
               Using <code>zookeeper-shell.sh</code>, run <code>rmr /controller</code> so that one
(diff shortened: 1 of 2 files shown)


lzyLuke (approved):

38/ops.html
```diff
@@ -4105,7 +4105,10 @@ <h3>Reverting to ZooKeeper mode During the Migration</h3>
             </li>
             <li>
               Using <code>zookeeper-shell.sh</code>, run <code>rmr /controller</code> so that one
-              of the brokers can become the new old-style controller.
+              of the brokers can become the new old-style controller. Additionally, run
+              <code>get /migration</code> followed by <code>rmr /migration</code> to clear the
+              migration state from ZooKeeper. This will allow you to re-attempt the migration
+              in the future. The data read from "/migration" can be useful for debugging.
```
showuon, line 4111: We also need to update it in L4147 (`Migrating brokers to KRaft` row)

chia7712: oh, I don't notice this comment. https://github.com/apache/kafka/pull/16770 is merged. I will sync the changes back to kafka repo if it needs

ppatierno: I have commented the corresponding JIRA to understand in which state the user rollbacks the cluster to restart the migration because I tried this scenario more times without deleting the /migration znode and it worked fine.

mumrah: Thanks @showuon, added the second mention. I wonder why we have this duplication? 🤔

chia7712: > I wonder why we have this duplication?

it seems we list all directions for reverting for each migration phase. That is a bit verbose, but they are useful. However, it is weird to use "table format" to display so much texts. Maybe we should to a bit refactor for it. Also, the same direction can have its hyperlink to be shared in different migration phase.

chia7712: I file jira for it (https://issues.apache.org/jira/browse/KAFKA-17271)

mjsax:
@mumrah -- can this PR be merged?

ppatierno:
It's still not clear to me why this removal is needed. I ran migration then rollback then migration and so on several times without the need for deleting such znode.

PR #625, opened 2024-08-18
MINOR: inter.broker.protocol.version is not clearly in KRaft and Zookeeper mode

m1a2st:
In Zookeeper mode, use `inter.broker.protocol.version' configuration to setting the MV, but it doesn't work in KRaft mode, thus I think the document should description more clearly on the config.


chia7712:
Could you please apply this fix to all versions, which using the argument. also please file a PR for kafka repo too.

chia7712:
related to https://github.com/apache/kafka/pull/16910

chia7712 (approved):
LGTM

PR #631, opened 2024-09-17
MINOR: add Colin P. McCabe key

cmccabe:
Add a new code signing key for Colin P. McCabe.


mumrah (approved):
LGTM

I could not find this key in OpenPGP key server: https://keys.openpgp.org/search?q=cmccabe%40apache.org

Can you confirm if the key made it there?

cmccabe:
I see it on `keyserver.ubuntu.com` but not on `keys.openpgp.org`. I will add it to the latter now.

cmccabe:
Apparently you have to have a blank line after
```
-----BEGIN PGP PUBLIC KEY BLOCK-----
```
or some scripts will fail. I updated the PR.

PR #634, opened 2024-10-08
MINOR Add upgrade instructiosn for 38 and 39

jlprat:
During the last 38 version we (actually it was me) forgot to copy over the instructions for a rolling upgrade.
I'm fixing it for 38 and also for 39 as both have are missing it.


chia7712 (approved):
LGTM

PR #635, opened 2024-10-16
[Docs] Update site for 3.8.1

jlprat:
Updating the site for 3.8.1.
As https://home.apache.org is now gone, we can't have a real preview of the site for patch versions as all changes would be visible directly.

### DIFF
--- a/38/js/templateData.js
+++ b/38/js/templateData.js
@@ -19,6 +19,6 @@ limitations under the License.
 var context={
     "version": "38",
     "dotVersion": "3.8",
-    "fullDotVersion": "3.8.0",
+    "fullDotVersion": "3.8.1-SNAPSHOT",
     "scalaVersion": "2.13"
 };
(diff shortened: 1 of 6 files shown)


38/js/templateData.js
```diff
@@ -19,6 +19,6 @@ limitations under the License.
 var context={
     "version": "38",
     "dotVersion": "3.8",
-    "fullDotVersion": "3.8.0",
+    "fullDotVersion": "3.8.1-SNAPSHOT",
```
showuon: Is this correct? Should be `3.8.1` only?

jlprat: I think it should be 3.8.1 yes. I'll fix it shortly

jlprat: fixed

showuon (approved):
LGTM!

mimaison:
Should we move 3.8.0 out of the supported releases?

jlprat:
I guess so. As 3.8.1 superseeded 3.8.0. I'll push a PR changing it

jlprat:
Here is the PR: https://github.com/apache/kafka-site/pull/638

PR #648, opened 2024-11-19
MINOR: Fix the leak "unknown" group.coordinator.rebalance.protocols on document

m1a2st:
Although "unknown" group.coordinator.rebalance.protocols type is a valid string in internal kafka system, but the common user should not config this "unknown" type, so we should remove it from the document.


m1a2st:
This PR is associate with [PR](https://github.com/apache/kafka/pull/17834).

chia7712 (approved):

PR #655, opened 2024-12-18
MINOR: update download links

mjsax:
- https://www.apache.org/legal/release-policy.html#release-announcements
- https://infra.apache.org/release-download-pages.html

### DIFF
--- a/downloads.html
+++ b/downloads.html
@@ -29,11 +29,13 @@
                     Docker Native image: <a href="https://hub.docker.com/layers/apache/kafka-native/3.9.0/images/sha256-ecf23c833391b00803876332b8642e261e675237e133cba8bc6a59ab8292a4a6">apache/kafka-native:3.9.0</a>.
                 </li>
                 <li>
+                    <!-- Note that source/binary downloads should point to `dlcdn.apache.org` but checksum files must point to `downloads.apache.org` -->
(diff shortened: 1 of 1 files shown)


jlprat (approved):
LGTM! Thanks @mjsax 
At first I though you missed all checksums and signature links but then I realize that infra specifies that those should be pointing to `downloads`:
> All links to checksums, detached signatures and public keys must reference the Apache Distribution Directory ([downloads.apache.org](https://downloads.apache.org/)) and use https:// (SSL). For example: https://downloads.apache.org/httpd/KEYS.

mjsax:
Yes, I am just not sure what to do with the release notes? I did reply on the dev mailing list to Craig Russell and hope he can clarify.

jlprat:
My guess would be just binaries/sources only because of the volume, but this is just my educated guess.

clr-apache:
Yes, the changes from downloads to dlcdn are correct. The release notes should also refer to dlcdn as well. The reason for using downloads is to make sure that the small checksums and signatures are not intercepted. But the larger artifacts need to be load balanced by using dlcdn.
Thanks for taking care of this.

mjsax:
Thanks @clr-apache -- pushed an update to this PR.

Also added a few more improvements \cc @jlprat

jlprat (approved):
LGTM! Thanks! And good thing to add the comments for future versions to remember to use the right domains for each type of file

PR #671, opened 2025-03-13
Revert "MINOR: Update default docs to point to the 4.0 release docs

frankvicky:
This reverts commit f4886c8fa2776f5e95c61981ebaba66fb76b81b0.
We should not modify the pointer of default docs before release.


frankvicky:
@dajac @showuon @chia7712 
PTAL

dajac (approved):

PR #674, opened 2025-03-18
MINOR: Small edits for 4.0

dajac:


ijuma (approved):

frankvicky (approved):
This change makes sense to me.
We only have one binary file now.

PR #678, opened 2025-03-20
MINOR: Show Committer pictures on small screens

jlprat:
Before the change:
Screen Shot 2025-03-20 at 17 21 01
After the change:
Screen Shot 2025-03-20 at 17 21 19

It's not perfect, but with the space given and not making a full fledged change on the table format, it's the best I could do with minimal effort.

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -1680,6 +1680,10 @@ nav .btn {
     h4 {
         font-size: 1.6rem;
     }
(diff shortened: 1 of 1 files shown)


chia7712 (approved):

jlprat:
Thanks @chia7712.
This thing was annoying me too much for some time now.

PR #684, opened 2025-04-16
Added missing `/migration` znode deletion on rollback KRaft migration

ppatierno:
The table provided in the section "Reverting to ZooKeeper mode During the Migration" of the KRaft migration guide is written in a way that each line is self-consistent (or this is my understanding).
It means that given the phase you are (so looking at the column "Final Migration Section Completed"), you refer to one line of the table and follow the instruction showed in the "Directions for Reverting" column (+ "Notes" one).
When in "Migrating brokers to KRaft" phase, the steps are missing the deletion of the `/migration` znode and this is fixed by this PR.

### DIFF
--- a/39/ops.html
+++ b/39/ops.html
@@ -4292,7 +4292,10 @@ listeners=CONTROLLER://:9093
             </li>
             <li>
               Using <code>zookeeper-shell.sh</code>, run <code>delete /controller</code> so that one
-              of the brokers can become the new old-style controller.
+              of the brokers can become the new old-style controller. Additionally, run
+              <code>get /migration</code> followed by <code>delete /migration</code> to clear the
+              migration state from ZooKeeper. This will allow you to re-attempt the migration
+              in the future. The data read from "/migration" can be useful for debugging.
             </li>
             <li>
               On each broker, remove the <code>zookeeper.metadata.migration.enable</code>,


ppatierno:
@mimaison can you take a look at this one please?

cmccabe:
@mumrah @mimaison @ppatierno the states are in the wrong order here. Fix here : https://github.com/apache/kafka/pull/19499

fvaleri:
@ppatierno thanks for the update. Wouldn't be better to put the `get /migration` part in the notes?

ppatierno:
@fvaleri it's not "just" a note, it's a clear statement of what it's needed to do. If you see, the same statement is present in the previous phase/line and in the same column.

fvaleri:
Well I still see the notes as a better fit for a command that may or may not be useful. That said, let's leave it like this for consistency.

fvaleri (approved):

mimaison (approved):
LGTM

mimaison:
Can you also open a PR against `apache/kafka:3.9` (https://github.com/apache/kafka/blob/3.9/docs/ops.html) with the same changes? Thanks

ppatierno:
@mimaison sorry for the delay! I missed your comment, here's the PR! https://github.com/apache/kafka/pull/19953

PR #719, opened 2025-09-03
MINOR: Add 4.0.1 docs

clolov:


clolov:
I had a quick check, because I was surprised to see the protocol_messages.html being included. As far as I can see the additions (I spot-checked some of them), have been present since December 2024 and are in the source code. As such, I think the current PR is correct, but I am happy to hear other opinions on the subject because I might be overlooking something simple

40/documentation.html
```diff
@@ -30,7 +30,7 @@
     <!--#include virtual="toc.html" -->
   </div>
   <div class="right">
-    <!--#include virtual="../includes/_docs_banner.htm" -->
+    <!--//#include virtual="../includes/_docs_banner.htm" -->
```
mimaison: We don't want this change. We need to show the docs banner on 4.0 as 4.1 is now the latest docs.

40/streams/developer-guide/datatypes.html
```diff
@@ -103,7 +103,7 @@ <h3>Primitive and basic types<a class="headerlink" href="#primitive-and-basic-ty
         <pre class="line-numbers"><code class="language-xml">&lt;dependency&gt;
     &lt;groupId&gt;org.apache.kafka&lt;/groupId&gt;
     &lt;artifactId&gt;kafka-clients&lt;/artifactId&gt;
-    &lt;version&gt;{{fullDotVersion}}&lt;/version&gt;
+    &lt;version&gt;2.8.0&lt;/version&gt;
```
mimaison: Why are we putting 2.8.0 here? I think we want `{{fullDotVersion}}` here

40/streams/upgrade-guide.html
```diff
@@ -224,7 +224,7 @@ <h3><a id="streams_api_changes_400" href="#streams_api_changes_400">Streams API
 
     <p>
         In order to reduce storage overhead and improve API usability, a new method in the Java and Scala APIs that accepts a BiFunction for foreign key extraction is introduced by
-        <a href="https://cwiki.apache.org/confluence/x/gIuMEw">KIP-1104</a>.
+        <a href="">KIP-1104</a>.
```
mimaison: This change breaks the link

40/generated/protocol_messages.html
mimaison, line 1: I wasn't expected changes here but that looks correct. Not sure what happened with `protocol_messages.html` for 4.0 but it looks like it's currently wrong.

40/streams/developer-guide/datatypes.html
clolov: I will have a deeper look into this and the one below. Your observation is correct, but I am uncertain how this got auto-generated

mimaison (approved):
LGTM

PR #721, opened 2025-09-08
MINOR: update docs for KAFKA-19668

mjsax:


bbejeck (approved):
LGTM

33/streams/developer-guide/dsl-api.html
```diff
@@ -3496,6 +3496,9 @@ <h5><a class="toc-backref" href="#id34">KTable-KTable Foreign-Key
                     </tr>
                     </tbody>
                 </table>
+
+                <p><b>CAUTION:</b> If you are using "merge repartition topics" optimization, it is not recommended to use <code>KStream#processValues</code> to avoid compatibility issues for future upgrades to newer versions of Kafka Streams.
+                    For more details, see the <a href="/40/documentation/streams/developer-guide/dsl-api.htm#transformers-removal-and-migration-to-processors">migration guide</a> in the Kafka Streams 4.0 docs.</p>
```
mjsax, line 3501: This update (same for other version up to, including, `3.8`) are only done here on `kafka-site.git`, but not in `kafka.git`. But should be ok, as we won't do any more 3.x bug-fix releases with high probability.

I did put this line into `3.9` though, as it's the last release in `3.9` series and might get more bug-fix releases, if people are hesitant to migrate to `4.0`.

PR #731, opened 2025-10-13
Minor: Add blog and downloads for 4.0.1

clolov:


downloads.html
```diff
@@ -13,68 +13,67 @@ <h1>Download</h1>
         
         <h2>Supported releases</h2>
 
-        <span id="4.1.0"></span>
-        <h3 class="download-version">4.1.0<a href="#4.1.0"><i class="fas fa-link " style="color:#053ce2"></i></a></h3>
+        <span id="4.0.1"></span>
```
mimaison: I wonder if keeping 4.1.0 at the top makes more sense. WDYT?

downloads.html
```diff
@@ -116,6 +115,38 @@ <h3 class="download-version">3.9.1<a href="#3.9.1"><i class="fas fa-link " style
 
         <h2>Archived releases</h2>
 
+            <span id="4.0.0"></span>
```
mimaison, line 118: For archive releases, update the links to use `archive.apache.org` instead of the CDN. See other releases for an example.

downloads.html
clolov: Agreed, moving back to the top!

downloads.html
clolov: Fair point! I hope I have addressed all links in the newest commit!

mimaison (approved):
LGTM

PR #733, opened 2025-10-26
KAFKA-19835: The Content-Security-Policy header must not be overridden

brandboat:
The Content-Security-Policy header must not be overridden.
There is now a standard way to add local exceptions to the CSP:
https://infra.apache.org/tools/csp.html

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -10,4 +10,6 @@ RewriteRule ^/?(\d+)/javadoc - [S=2]
 RewriteRule ^/?(\d+)/images/ - [S=1]
 RewriteCond $2 !=protocol
 RewriteRule ^/?(\d+)/([a-z]+)(\.html)? /$1/documentation#$2 [R=302,L,NE]
-Header set Content-Security-Policy "frame-src https://youtube.com https://www.youtube.com"
+
+# Adding YouTube for embedded videos to CSP header, more details please check https://infra.apache.org/tools/csp.html
+SetEnv CSP_PROJECT_DOMAINS "https://youtube.com https://www.youtube.com"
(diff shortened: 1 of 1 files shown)


brandboat:
quickstart local dev env screenshot:

image

sebbASF:
Note that the .htaccess file needs to document why the override is allowed.

brandboat:
@sebbASF, thanks for the comment! The reason we add youtube here is due to we use embedded youtube videos in QuickStart, KafkaStreams page, see https://kafka.apache.org/quickstart, https://kafka.apache.org/documentation/streams/
Without this, browsers will block these iframes and videos won't display.

sebbASF:
Your explanation covers why the override is needed.
However it does not cover why the override is allowed.
According to https://infra.apache.org/tools/csp.html

"Each additional host you add MUST have been pre-approved by VP Data Privacy ([privacy@apache.org](mailto:privacy@apache.org)), and SHOULD have an accompanying comment in the .htaccess file explaining why the CSP is changed and where permission was obtained."

brandboat:
> Each additional host you add MUST have been pre-approved by VP Data Privacy ([privacy@apache.org](mailto:privacy@apache.org))

@sebbASF, thanks again for the comment! Could you please explain how this approval process is supposed to be done?
I don’t seem to have access to the email threads at `privacy@apache.org`, so I’m wondering whether this step needs to be handled by an Apache member or a project lead.
Sorry if this is a basic question — I’m just a casual contributor and not very familiar with the internal process. c.c. @chia7712

sebbASF:
Such questions should be directed to the Kafka PMC in the first instance

mjsax:
Thanks for raising this @sebbASF -- I just stubbled into this PR. Not familiar with the process here, but yes, we can take it to the PMC mailing list. @chia7712, do you want to own it, given that you did approve the original PR modifying `.htaccess` file back in the days: https://github.com/apache/kafka-site/commit/8d6014f1b4811a0e7d61352b34faedadd85db77c

chia7712:
> do you want to own it

sure, no problem 😃

mumrah:
As per KIP-1133, we have converted our docs over to Markdown. This repo has also changed its base branch to `markdown`. If you still want this change applied, please reopen a PR using Markdown against the `markdown` branch. Thanks!

PR #747, opened 2025-11-15
KAFKA-19751: KAFKA-19751: add Java 21, 23 to 3.9 supported version

FrankYang0529:
https://issues.apache.org/jira/browse/KAFKA-19751

reviewers: Chia-Ping Tsai <chia7712@gmail.com>

### DIFF
--- a/39/ops.html
+++ b/39/ops.html
@@ -1249,10 +1249,10 @@ queued.max.requests=[number of concurrent requests]</code></pre>
 
   <h3 class="anchor-heading"><a id="java" class="anchor-link"></a><a href="#java">6.6 Java Version</a></h3>
 
-  Java 8, Java 11, and Java 17 are supported.
+  Java 8, Java 11, Java 17, Java 21 and Java 23 are supported.
   <p>
   Note that Java 8 support project-wide has been deprecated since Apache Kafka 3.0 and Java 11 support for the broker and tools
-  has been deprecated since Apache Kafka 3.7. Both will be removed in Apache Kafka 4.0.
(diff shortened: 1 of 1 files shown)


FrankYang0529:
Preview:
Screenshot 2025-11-15 at 1 54 24 PM

chia7712 (approved):

39/ops.html
```diff
@@ -1249,10 +1249,10 @@ <h4 class="anchor-heading"><a id="prodconfig" class="anchor-link"></a><a href="#
 
   <h3 class="anchor-heading"><a id="java" class="anchor-link"></a><a href="#java">6.6 Java Version</a></h3>
 
-  Java 8, Java 11, and Java 17 are supported.
+  Java 8, Java 11, Java 17, Java 21 and Java 23 are supported.
```
ijuma, line 1252: Not sure we want to include Java 23, which is a non LTS version that is not supported by anyone else.

chia7712: That's a good question. Since the 3.9 branch will be a specific version with potentially many patch releases, its lifecycle may be inconsistent with the JDK EOL schedule. 

Perhaps we should state that JDK 23 is used by the CI but is not officially supported by Apache Kafka due to its EOL status. We can adopt this strategy again when JDK 8 reaches its EOL.

FrankYang0529: JDK 23 is EOL in March 2025. We can mention it's one of JDK version which is used in CI and it's not officially supported. I will create another PR for this.

FrankYang0529: https://github.com/apache/kafka-site/pull/748

PR #766, opened 2025-12-23
KAFKA-17862: Update Producer Config `delivery.timeout.ms`

gongxuanzhang:
see https://github.com/apache/kafka/pull/19489

After this PR merge, we should update producer config `delivery.timeout.ms` desc

The figure below shows the behavior after the change.

image


chia7712:
With the new docs workflow, we don't need to update the site. Instead, please update the docs in the Kafka repo directly.

PR #774, opened 2026-01-04
fix: malformed GitHub URL with double slash and Update feedback link from GitHub Issues to Apache JIRA

ArmanTaheriGhaleTaki:
# Summary
This PR fixes a malformed GitHub URL that contained an erroneous double slash (//) in the path. The incorrect URL structure was causing potential resolution issues and follows best practices for clean, canonical URLs.

## Changes Made
Fixed URL path: Changed https://github.com/apache/kafka-site//tree/markdown/
image

 to https://github.com/apache/kafka-site/tree/markdown/...
 
image

Removed duplicate slash: Eliminated the redundant // that appeared before tree in the URL path

## and there is another change 
The kafka-site repository has disabled its GitHub Issues section, making the original link non-functional for users trying to submit feedback. All Apache Kafka project issues are now centrally managed through the official Apache JIRA instance, which is the standard issue tracking system for Apache Software Foundation projects.

### DIFF
--- a/hugo.yaml
+++ b/hugo.yaml
@@ -79,7 +79,7 @@ outputs:
 #     - name: doc
 #       url: /latest/docs/
 #       weight: 1
-      
+
 #     - name: "Documentation"
 #       url: "/latest/docs/"
 #       weight: 1
@@ -204,7 +204,7 @@ menu:
       parent: "Community"
       url: "/community/contact/"
       weight: 47
-    
+
     # Apache menu
     - name: "Apache"
       url: "https://www.apache.org/"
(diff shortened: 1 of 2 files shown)


hugo.yaml
```diff
@@ -437,9 +437,9 @@ params:
       enable: false
       # The responses that the user sees after clicking "yes" (the page was helpful) or "no" (the page was not helpful).
       'yes': >-
-        Glad to hear it! Please <a href="https://github.com/apache/kafka-site/issues/new">tell us how we can improve</a>.
+        Glad to hear it! Please <a href="https://issues.apache.org/jira/browse/KAFKA">tell us how we can improve</a>.
```
mimaison, line 440: I think this feature is disabled as the website does not have Google Analytics. So we may just be able to remove this configuration.

mimaison: We have a ` Create documentation issue` link on all pages that has links to Github issues. This would need updating but I'm not sure where it's configured though.

mimaison (commented):
Thanks for the PR. I'm not sure it's addressing the issue properly.

PR #779, opened 2026-01-09
Add .htaccess file

tisonkun:
cc @mumrah 

Then we need not manually update the one in production like https://github.com/apache/kafka-site/commit/106357750f424c22370eebe2e1066534ffe34c36.

See also https://lists.apache.org/thread/npx5d5w2m0ocs7mv71lxslz0pvbl34o2 for why we need the redirect rule.

### DIFF
--- a/static/.htaccess
+++ b/static/.htaccess
@@ -1,0 +1,5 @@
+# CSP Settings
+Header set Content-Security-Policy "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; script-src 'self' 'unsafe-inline' https://unpkg.com https://code.jquery.com"
+
+# Backward compatibility redirections
+Redirect 301 /KEYS https://downloads.apache.org/kafka/KEYS


tisonkun:
```
#15 [auth] apache/kafka-site:pull,push token for ghcr.io
#15 DONE 0.0s

#14 exporting to image
#14 pushing layers 0.7s done
#14 ERROR: failed to push ghcr.io/apache/kafka-site:pr-779: denied: installation not allowed to Write organization package
```

Seems due to I cannot push a branch under this repo :P

tisonkun:
Can be duplicated to https://github.com/apache/kafka-site/pull/782. Either one is OK to me.

static/.htaccess
```diff
@@ -0,0 +1,5 @@
+# CSP Settings
```
chia7712, line 1: Should we need to add `RewriteEngine On`?

static/.htaccess
```diff
@@ -0,0 +1,5 @@
...
+Header set Content-Security-Policy "style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; script-src 'self' 'unsafe-inline' https://unpkg.com https://code.jquery.com"
```
chia7712, line 2: not sure whether this is required. could you share more details?

tisonkun: I guess it's about ASF's CSP requirement.

This is from the online version .htaccess.

tisonkun:
@chia7712 Thanks for your review.

I think the Kafka community can properly handle this issue and since I see no active response on merging my patch, I'd close this.

Perhaps you can follow up with https://github.com/apache/kafka-site/pull/782. I have no preference, but cannot allocate more time to a diverging ticket.

PR #830, opened 2026-03-14
KAFKA-20049: Publish Hugo base image under Apache namespace

Owen-CH-Leung:
This PR removes the short-term dependency on the personal `hvishwanath/hugo` image namespace and moves the repo toward an Apache-owned GHCR image.

Changes:
  - update the Hugo base image reference in `Makefile` to `ghcr.io/apache/kafka-site/hugo:v0.123.7-ext-multiplatform`
  - add a new GitHub Actions workflow to build and publish that Hugo base image from `Dockerfile.multiplatform`
  - update the Docker image workflow to use the Apache-owned image, with a local build fallback if the image is not published yet
  - update the README to reflect the new build flow

 This is intended as the short-term fix only: preserve the current custom Hugo image behavior, but stop relying on a personal image. 

Longer term, we should evaluate replacing this custom Kafka-owned Hugo image with the official Hugo image from the Hugo project, moving any repo-specific bootstrap steps into CI as needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,10 +3,8 @@ OUTPUT_DIR := output
 HUGO_VERSION := 0.123.7
 HUGO_BASE_IMAGE := ghcr.io/apache/kafka-site/hugo:v$(HUGO_VERSION)-ext-multiplatform
 DOCKER_IMAGE := $(HUGO_BASE_IMAGE)
-#PROD_IMAGE := hvishwanath/kafka-site-md:1.2.0
-PROD_IMAGE := us-west1-docker.pkg.dev/play-394201/kafka-site-md/kafka-site-md:1.6.0
 
-.PHONY: build serve clean docker-image ensure-hugo-image hugo-base-multi-platform prod-image prod-run buildx-setup ghcr-prod-image
+.PHONY: build serve clean docker-image ensure-hugo-image hugo-base-multi-platform buildx-setup ghcr-prod-image
 
 # Setup buildx for multi-arch builds
 buildx-setup:
@@ -49,20 +47,6 @@ serve: ensure-hugo-image
 		--buildDrafts \
 		--buildFuture
 
-# Build production Nginx image for multiple architectures
-prod-image: build buildx-setup
-	docker buildx build \
-		--platform linux/amd64,linux/arm64 \
-		--tag $(PROD_IMAGE) \
-		--file Dockerfile.prod \
-		--push \
-		.
-
-# Run production image locally
-prod-run: prod-image
-	docker pull $(PROD_IMAGE)
-	docker run --rm -p 8080:80 $(PROD_IMAGE)
-
 # Build and push production image to GHCR
 ghcr-prod-image: build buildx-setup
 	docker buildx build \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,11 +1,12 @@
 # Hugo configuration
 OUTPUT_DIR := output
-HUGO_BASE_IMAGE := hvishwanath/hugo:v0.123.7-ext-multiplatform
+HUGO_VERSION := 0.123.7
+HUGO_BASE_IMAGE := ghcr.io/apache/kafka-site/hugo:v$(HUGO_VERSION)-ext-multiplatform
 DOCKER_IMAGE := $(HUGO_BASE_IMAGE)
 #PROD_IMAGE := hvishwanath/kafka-site-md:1.2.0
```
mimaison: Can we remove this comment too?

Makefile
```diff
@@ -1,11 +1,12 @@
...
 PROD_IMAGE := us-west1-docker.pkg.dev/play-394201/kafka-site-md/kafka-site-md:1.6.0
```
mimaison: What is `PROD_IMAGE`? Who owns it?

mimaison (commented):
Thanks for the PR. I left a question

Makefile
Owen-CH-Leung: This was pre-existing in markdown HEAD and not introduced by this PR. I wanted to flag it and ask for your comment before touching it.

`PROD_IMAGE` points to a personal GCP Artifact Registry (`us-west1-docker.pkg.dev/play-394201/...`) and is used only by the `prod-image` and `prod-run` Makefile targets — neither of which is called by any CI workflow. These appear to be local development convenience targets for building and running the nginx+site image locally.

 Should we remove `PROD_IMAGE`, `prod-image`, and `prod-run` from the Makefile since they depend on a personal registry that no other contributor can push to? Or is there a legitimate use case for these targets that we should preserve with an Apache-owned registry instead?

mimaison: We shouldn't have or depend on any privately owned artifacts. If this is not used let's remove it. If it's used we need to replace it with something Apache owns.

Owen-CH-Leung: Agree. I've removed the unused targets in Makefile. There's no dependency on private artifact registry anymore

mimaison:
The changes seem fine to me but I'm not an expert with the website pipeline.
@mumrah @hvishwanath can you also take a look?

hvishwanath:
Thanks @Owen-CH-Leung for cleaning this up. This LGTM!

Owen-CH-Leung:
Thanks @hvishwanath. Shall we merge the MR ?

mimaison:
> Longer term, we should evaluate replacing this custom Kafka-owned Hugo image with the official Hugo image from the Hugo project, moving any repo-specific bootstrap steps into CI as needed.

@Owen-CH-Leung Do you have an idea how the effort this would take?

Owen-CH-Leung:
> > Longer term, we should evaluate replacing this custom Kafka-owned Hugo image with the official Hugo image from the Hugo project, moving any repo-specific bootstrap steps into CI as needed.
> 
> @Owen-CH-Leung Do you have an idea how the effort this would take?

@mimaison  I've prepared another PR [here](https://github.com/apache/kafka-site/pull/885) that moves away from the custom image. Feel free to review and comment

PR #849, opened 2026-04-18
KAFKA-20463: Auto-deploy to asf-site on PR merge

Owen-CH-Leung:
Previously, merging a PR only deployed to asf-staging because `inputs.deploy-to-asf-site` is null for push events. The live site required a manual workflow_dispatch with the checkbox enabled.                                                                                                                                                           
                                                                                                                                                                                                                           
Now push events deploy directly to asf-site. Manual runs without the checkbox still target asf-staging for testing.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -51,7 +51,7 @@ jobs:
         GITHUB_CONTEXT: ${{ toJson(github) }}
     - name: Configure
       id: configure
-      run: echo "ref=${{ (github.event_name == 'push' || inputs.deploy-to-asf-site) && 'asf-site' || 'asf-staging' }}" >> "$GITHUB_OUTPUT"
+      run: echo "ref=${{ inputs.deploy-to-asf-site && 'asf-site' || 'asf-staging' }}" >> "$GITHUB_OUTPUT"
     - name: Checkout
       uses: actions/checkout@v5
       with:
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -44,6 +44,8 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     needs: build-html
+    permissions:
+      pull-requests: write
     steps:
     - name: Env
       run: printenv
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-deploy.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         GITHUB_CONTEXT: ${{ toJson(github) }}
     - name: Configure
       id: configure
-      run: echo "ref=${{ inputs.deploy-to-asf-site && 'asf-site' || 'asf-staging' }}" >> "$GITHUB_OUTPUT"
+      run: echo "ref=${{ (github.event_name == 'push' || inputs.deploy-to-asf-site) && 'asf-site' || 'asf-staging' }}" >> "$GITHUB_OUTPUT"
```
chia7712: I'm not sure if this is safe for us, as it could expose errors directly to our readers. Also, it essentially renders `asf-staging` useless, meaning we won't have the chance to preview or debug the website there beforehand

ping @mimaison WDYT?

mimaison: Yeah actually it seems this is a "feature" as it's described in https://github.com/apache/kafka-site#updating-the-website.

However I find the process a bit confusing, and I wasn't quite aware of it. I guess I'm not the only one as I've noticed the website being out of date a few times recently.

Owen-CH-Leung: How about this workflow ? 
 
- We push to `asf-staging` automatically on PR merge
- We add a separate scheduled/auto job that promotes `asf-staging` to `asf-site` that runs twice a week

chia7712: > We add a separate scheduled/auto job that promotes asf-staging to asf-site that runs twice a week

sounds good :)

mimaison: +1, thanks

Owen-CH-Leung: Added a new CI scheduled job that promotes `asf-staging` to `asf-site` that runs every Monday & Thursday 02:00 UTC. Can you review agian ?

mjsax: I think the goal of targeting staging is to allow us to review the changes "live". If we could have a way to deploy the web-page on a PR, we would still have the safeguard in place? Not sure if this would be (easily) possible.

chia7712: > Not sure if this would be (easily) possible.

Another simple approach is a post-commit check. After deploying to the staging site, the CI can leave a comment on the PR to remind the author to double-check the results. We can even include a note like: 'This will go live in 24 hours (or on the next schedule). Please file a follow-up PR if you spot any issues on staging.

muralibasani: May be this be an option https://github.com/marketplace/actions/build-deploy-to-github-pages-and-deploy-pr-preview ? 
On every commit we get a preview on GH Pages.

Owen-CH-Leung: I like @chia7712 idea. I have added a new step to the deploy job that posts a comment on the merged PR once staging is updated. It only fires on push-triggered runs (not manual dispatches) and silently skips if there's no associated PR. The comment links to the staging site and notes the Mon/Thu promotion schedule.

PR #860, opened 2026-05-12
KAFKA-20488: KRaft migration rollback causes offline partitions when KRaft epoch exceeds ZK controller epoch

mwesterby:
### Description 
- Updates the ZooKeeper to KRaft migration docs with the controller epoch issue which may occur when reverting back to ZK mode.
- Adds steps on how to verify the ZK controller epoch is greater than the KRaft controller epoch prior to reverting back to ZK mode. Additionally adds the workaround on what to do if this is not the case.
- I've reworded the note around performing the "`zookeeper-shell.sh` step **quickly**" to be "perform the `delete /controller` step **quickly**", as these new steps also include using a zookeeper shell.

### Jira
https://issues.apache.org/jira/browse/KAFKA-20488


showuon (approved):
LGTM!

PR #864, opened 2026-05-20
MINOR: Add 4.3.0 to Downloads page

mimaison:


mimaison:
@AndrewJSchofield @clolov Can you take a look? Thanks

jlprat:
@mimaison Shouldn't we move the 3.9.x to archive now that 4.3.0 is released?

jlprat (approved):
The section for 4.3.0 itself looks good to me. My only question as I shared in the other comment is if we move 3.9.x to the archive section or not.

mimaison:
Yes I hesitated whether we should archive 3.9 or not. The 3.9.2 release is fairly recent. I'll merge this PR so I can announce 4.3.0 and we can then decide what to do with 3.9.

PR #879, opened 2026-06-15
Add back negation that was lost in rewrite

TitouanT:
Before version 40 the text was:
```
It is generally _not_ advisable to run a _single_ Kafka cluster that spans multiple datacenters over a high-latency link. This will incur very high replication latency both for Kafka writes and ZooKeeper writes, and neither Kafka nor ZooKeeper will remain available in all locations if the network between locations is unavailable.
```

The negation got lost when removing Zookeeper from that paragraph.


mimaison (approved):
LGTM, thanks for the PR.

Can you also open a PR against the main Kafka repo? The file is https://github.com/apache/kafka/blob/trunk/docs/operations/datacenters.md

PR #882, opened 2026-06-16
MINOR: Fix typo in config name

morgando:
`listener.security.property.map` should be `listener.security.protocol.map`


showuon (approved):

showuon:
Nice catch! Thank you @morgando !

PR #885, opened 2026-06-24
MINOR: Add a build image based on the official Hugo release binary

Owen-CH-Leung:
## Summary

Add a local build image `Dockerfile.hugo-builder` that installs the official Hugo extended release binary from the Hugo project, alongside the other required binaries such as the `Node.js`, `PostCSS` and Go toolchains . 

Unlike the current build image, it does not depend on any externally-hosted or personal image (e.g. hvishwanath/hugo).

This is added as a separate, opt-in path, so no CI behaviour or deployment changes in this PR:

- CI: the "Build and Deploy Site" workflow runs `make build`, which is left unchanged and still uses the current image. The published website is not affected.
- Local: `make build-official` builds the new image from the official binary and renders the site with it. Its output is byte-for-byte identical to `make build`

The goal is to validate the official-binary image first; switching the build and CI over to it can follow in a later change.

### DIFF
--- a/Dockerfile.hugo-builder
+++ b/Dockerfile.hugo-builder
@@ -1,0 +1,32 @@
+ARG HUGO_VERSION=0.123.7
+
+FROM alpine:3.21
+
+RUN apk add --no-cache \
+    bash \
+    ca-certificates \
+    curl \
+    git \
+    go \
+    libc6-compat \
+    nodejs \
+    npm
+
+ARG HUGO_VERSION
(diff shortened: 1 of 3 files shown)


Owen-CH-Leung:
Instead of using the official hugo image, I've built the image from alpine and download the hugo binary **v0.123.7**. This is because the official image isn't published for 0.123.7 (its oldest tags are 0.135)

PR #889, opened 2026-06-25
MINOR: Updates for release

bbejeck:
Updates the correct version on the APIs page and adds upgrade to section for `4.3.1`

### DIFF
--- a/content/en/43/getting-started/upgrade.md
+++ b/content/en/43/getting-started/upgrade.md
@@ -26,6 +26,15 @@ type: docs
 -->
 
 
+## Upgrading to 4.3.1
+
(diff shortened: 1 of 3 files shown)


frankvicky (approved):

content/en/43/getting-started/upgrade.md
```diff
@@ -26,6 +26,15 @@ type: docs
 -->
 
 
+## Upgrading to 4.3.1
```
chia7712, line 29: This section should be synced to the kafka repo as well, otherwise it will be overwritten by the next release. I will file a patch to address it.

PR #894, opened 2026-07-11
MINOR: Sync doc description fix from apache/kafka#22605 to 4.3.X pages

mingyen066:
apache/kafka#22605 fixed duplicate/blank section descriptions in the
docs/ source, but kafka-site doesn't auto-sync from kafka trunk — the
already-published 4.3.X pages here were untouched. This backports the
same description text to the 82 matching files under content/en/43/, so
the live site gets the fix now instead of waiting for the next release.

Before:

After:

### DIFF
--- a/content/en/0100/configuration/broker-configs.md
+++ b/content/en/0100/configuration/broker-configs.md
@@ -1,6 +1,6 @@
 ---
 title: Broker Configs
-description: Broker Configs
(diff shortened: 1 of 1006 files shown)


mingyen066:
Need to solve conflict with https://github.com/apache/kafka/pull/22605

PR #895, opened 2026-07-11
KAFKA-20009: Sync doc description fix from kafka trunk to active pages

mingyen066:
apache/kafka#22605 fixed duplicate/blank section descriptions in the
docs/ source, but kafka-site doesn't auto-sync from trunk, so the
already-published pages were untouched. Backports the same description
text to the three currently supported release lines (4.1.X, 4.2.X,
4.3.X); archived lines are left as is.


mingyen066:
don't need to backport

